PR #2, opened 2025-10-27
restructure README.md to follow python layout

gdams:
I think Copilot might have hallucinated about some of the implementation but I'll let @qmuntal confirm


qmuntal (commented):
Most of the samples won't work, but it's better than before :)

PR #7, opened 2025-11-11
Fixes and example enhancement

griffinbird:
Enhanced basic chat example to illustrate tool calling

Fix multiple tool call handling in OpenAI adapter

- Fix funcCall to return FunctionResultContent for all error cases
- Split tool result messages to send one message per tool call ID
- Ensure each tool_call_id gets its own response message per Azure OpenAI API requirements

This resolves 400 Bad Request errors when LLM makes multiple tool calls simultaneously.
The OpenAI adapter now correctly creates separate tool messages for each FunctionResultContent instead of bundling them into a single message with only the last call ID.

### DIFF
--- a/go/agent/agent.go
+++ b/go/agent/agent.go
@@ -389,10 +389,10 @@ func runToolCalls(ctx context.Context, options *RunOptions, contents ...Content)
 }
 
 // funcCall executes a function tool call.
-func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallContent) (ct Content) {
+func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallContent) Content {
 	if toolCall.Error != nil {
 		// If there was an error parsing the tool call, return the error.
-		return &FunctionCallContent{
+		return &FunctionResultContent{
 			CallID: toolCall.CallID,
 			Error:  toolCall.Error,
 		}
@@ -411,7 +411,7 @@ func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallCont
 	}
 
 	if found == nil {
-		return &FunctionCallContent{
+		return &FunctionResultContent{
 			CallID: toolCall.CallID,
 			Error:  fmt.Errorf("tool not found: %s", toolCall.Name),
 		}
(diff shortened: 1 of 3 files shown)


go/openai/chat.go
```diff
@@ -160,7 +160,23 @@ func (a *client) buildCompletionParams(options *agent.RunOptions, messages ...*a
 		Messages: make([]openai.ChatCompletionMessageParamUnion, 0, len(messages)+1),
 	}
 	for _, msg := range messages {
-		params.Messages = append(params.Messages, buildMessageParam(msg))
+		// Special handling for tool messages: each tool result needs its own message
```
qmuntal: Good catch. I like all the message handling logic to be in `buildMessageParam`, though. Could you update that function signature to returns a slice of `openai.ChatCompletionMessageParamUnion` instead and then build the multiple messages per tool call in there?

go/agent/agent.go
```diff
@@ -420,30 +420,35 @@ func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallCont
...
-				Error:  fmt.Errorf("tool execution panic: %v", err),
+				panicErr = fmt.Errorf("%v", r)
 			}
 		}
 	}()
 
 	// Execute the tool
 	result, err := found.Call(ctx, args)
+
+	if panicErr != nil {
```
qmuntal: I don't think this will work, we won't reach this point if the function call panicked, as the panic will be catch by the deferred recover and then immediately returned to the caller frame.

go/agent/agent.go
```diff
@@ -389,10 +389,10 @@ func runToolCalls(ctx context.Context, options *RunOptions, contents ...Content)
 }
 
 // funcCall executes a function tool call.
-func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallContent) (ct Content) {
+func funcCall(ctx context.Context, tools []tool.Tool, toolCall *FunctionCallContent) Content {
 	if toolCall.Error != nil {
 		// If there was an error parsing the tool call, return the error.
-		return &FunctionCallContent{
+		return &FunctionResultContent{
```
qmuntal: Is this fixing a bug you have experimented? AFAIU all previous returns of `FunctionCallContent` with an error (rather than `FunctionCallContent`) are correct.

Dotnet's [FunctionCallContent.Exception](https://learn.microsoft.com/en-us/dotnet/api/microsoft.semantickernel.functioncallcontent?view=semantic-kernel-dotnet) is documented to be set when the exception (our error) occurred while mapping original AI model function call to the model class, and that's what's happening here. Then, [FunctionResultContent.Exception](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.functionresultcontent?view=net-9.0-pp) is documented to be set if the actual function returned an error, so we should only use it when calling ` found.Call(ctx, args)`.

qmuntal (changes requested):

go/openai/chat.go
griffinbird: That's a good idea keeping with the single responsibility, and keeping buildCompletionParams for iterations and appends, and not handling any special cases.

go/agent/agent.go
griffinbird: Fixed - Wrapped found.Call() in an anonymous function with its own deferred recover()
When a panic occurs, it's caught and converted to an error in err
The anonymous function returns normally
The outer function then returns a single FunctionResultContent with either the panic error or the normal result

go/agent/agent.go
griffinbird: Sorry I misunderstood the difference between FunctionCallContent.Error and FunctionResultContent.Error.  I have reverted the changes.

qmuntal (approved):

PR #26, opened 2025-12-04
Make the agent package more idiomatic

qmuntal:
Refactor the `agent` package to be more idiomatic. The improvements in here are mostly taken from feedback form the last demo.

Notable changes:

- The `agent.Agent` interface has been simplified so it is easier to implement (see for example the A2A implementation).
- The `agent.Agent` callers are not expected to use an agent instance anymore, but a set of run functions defined in `agent`. This consolidates helper functions that where previously implemented on each agent implementation.
- The agent Run methods and functions no longer accepts an `agent.Context` (which has been removed). They now accept parameters that are easier to identify as a Go developer.

### DIFF
--- a/go/agent/a2aagent/a2a.go
+++ b/go/agent/a2aagent/a2a.go
@@ -136,12 +136,12 @@ func (a *Agent) sendMsg(ctx context.Context, options *agent.RunOptions, params *
 			yield(nil, err)
 			return
 		}
-		var contextID, taskID string
-		var update *agent.RunResponseUpdate
 		switch e := e.(type) {
 		case *a2a.Task:
-			contextID = e.ContextID
-			taskID = string(e.ID)
+			if err := a.updateThreadContextID(thread, e.ContextID, string(e.ID)); err != nil {
+				yield(nil, err)
+				return
+			}
 			now := time.Now()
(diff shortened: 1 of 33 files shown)


griffinbird:
@qmuntal is functional options a valid suggestion to keep in line with Go idiomatic patterns like in other Go SDKs?

- AWS SDK v2: config.LoadDefaultConfig(ctx, config.WithRegion("us-west-2"))
- Google Cloud: storage.NewClient(ctx, option.WithCredentials(...))

Easy for IDE autocomplete/discoverable, no mutation/clone issue, backwards compatible when adding new options.. I understand it's more verbose/boilerplate..  Thoughts?


```
func WithID (id string ) AgentOption {
    ...
}

func WithName (name string ) AgentOption {
    ...
}
```

etc

qmuntal:
> @qmuntal is functional options a valid suggestion to keep in line with Go idiomatic patterns like in other Go SDKs?

I personally dislike functional options, as I always struggle to find which functions among all the functions in a package are related to a given configuration/option.

Having said this, I don't close the door to using this pattern if it proves useful. In fact, I have tries this pattern to build RunOptions, but the result didn't convince me. Can give it another try latter on.

gdams (approved):

PR #36, opened 2025-12-10
fix: anthropic tool call

### LINKED ISSUE #35: Anthropic integration: tool_use.input passed as string instead of object
# Description
When using the Anthropic provider with tools, the `buildMessageParam` function passes `c.Arguments` (a JSON string) directly to `anthropic.NewToolUseBlock`, but the Anthropic SDK expects the `input` parameter to be a Go object (map/struct) that it will serialize.
This causes the Anthropic API to return:
400 Bad Request: "messages.1.content.1.tool_use.input: Input should be a valid dictionary"

## Location
`anthropic/chat.go`, line 370:
```go
case *message.FunctionCallContent:
    content = append(content, anthropic.NewToolUseBlock(c.CallID, c.Arguments, c.Name))
```

# Root Cause
- c.Arguments is a string containing JSON (e.g., "{\"repo\":\"owner/repo\"}")
- anthropic.NewToolUseBlock(id string, input any, name string) expects input to be a Go object that the SDK serializes
- The OpenAI integration works because OpenAI's API accepts string arguments directly
Suggested Fix
Unmarshal the JSON string into a map[string]any before passing to NewToolUseBlock:

```go
case *message.FunctionCallContent:
    var args map[string]any
    if c.Arguments != "" {
        if err := json.Unmarshal([]byte(c.Arguments), &args); err != nil {
            return anthropic.MessageParam{}, fmt.Errorf("failed to unmarshal tool arguments: %w", err)
        }
    }
    content = append(content, anthropic.NewToolUseBlock(c.CallID, args, c.Name))
```

# Steps to Reproduce
1. Create a chat agent using anthropic.NewChatAgent() with tools
2. Send a message that triggers tool use
3. The API returns 400 error when the agent tries to report tool usage back

# Environment
- agent-framework-go: latest main
- anthropic-sdk-go: v1.19.0
- Tested with Azure Anthropic endpoint

sozercan:
fixes #35  and more

### DIFF
--- a/anthropic/chat.go
+++ b/anthropic/chat.go
@@ -285,13 +285,28 @@ func (a *client) buildMessageParams(options *chatclient.ChatOptions, messages ..
 			var properties any
 			var required []string
 
-			// Extract schema details
-			switch schema := ft.Schema().(type) {
+			// Extract schema details - first convert to map[string]any if needed
+			schema := ft.Schema()
+			var schemaMap map[string]any
+
+			switch s := schema.(type) {
 			case map[string]any:
-				if props, ok := schema["properties"]; ok {
+				schemaMap = s
+			default:
+				// For *jsonschema.Schema or other types, marshal to JSON then unmarshal to map
+				if schema != nil {
+					jsonBytes, err := json.Marshal(schema)
+					if err == nil {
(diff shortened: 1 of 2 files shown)


qmuntal (approved):
Thanks!

PR #85, opened 2026-03-03
Simplify ResponseStream

qmuntal:
`ResponseStream`is clever in the way it automatically sets the streaming mode, But this behavior might be unexpected by users not reading the docs, and it diverges too much from the .NET/Python SDKs.


gdams (approved):
Much simpler!

PR #119, opened 2026-04-28
Add compaction support

qmuntal:
Ports https://github.com/microsoft/agent-framework/pull/4533.

Closes https://github.com/microsoft/go-lab/issues/338


gdams (approved):
LGTM although the `Parity gap: TurnIndex 0 is not explicitly protected` is worth confirming

qmuntal:
> The default `StateKey` in `NewContextProvider` is the fixed string `"CompactionProvider"`, while in .NET it defaults to the strategy type name (e.g., `"PipelineCompactionStrategy"`). The .NET default naturally avoids state-key collisions when multiple providers share a session; Go users with that setup need to set `SourceID`/`StateKey` explicitly. The field is documented and configurable, so this is a usage guidance difference rather than a breaking divergence.

Fixing 👍

PR #122, opened 2026-05-04
Improve Workflow execution API

qmuntal:
This PR refactors the Workflow execution API to simplify it and make it more similar to the .NET one.
It still needs some more love, though.

While here, add some more tests and expand workflow examples.

For https://github.com/microsoft/go-lab/issues/339


qmuntal:
> WithPendingRequestRepublish is public in Go, but the equivalent is internal in .NET.

That's fine. .NET exposes it through `WorkflowSession`, but Go doesn't have that abstraction layer.

gdams (approved):

PR #124, opened 2026-05-05
Unify agent workflow responses as output events

qmuntal:
## Summary
- remove workflow ResponseUpdateEvent and ResponseEvent in favor of OutputEvent payloads
- emit hosted-agent response updates and aggregated responses as OutputEvent values
- update workflow provider, examples, and tests
- port https://github.com/microsoft/agent-framework/pull/3441

## Tests
- go test ./...


qmuntal:
Agent response events are a bit more difficult to consume with this change, as they need one additional type cast. On the other hand, there is now no possible confusion about which event will be emitted when there is an agent response. It is not possible to achieve 100% parity with .NET due to the lack of inheritance, so I've picked the side of consistency and clarity over verbosity.

gdams (approved):

PR #137, opened 2026-05-06
Add graph validation for .NET/Python parity and housekeeping updates

gdams:
## Summary

Adds graph validation checks to the workflow builder for parity with the .NET and Python SDKs, along with housekeeping updates.

### Graph Validation
- **Type compatibility validation** (build error): At build time, creates temporary executor instances to inspect their routers and verifies that source output types overlap with target input types for each edge. Executors with catch-all handlers or no declared types are gracefully skipped.
- **Output executor validation** (build error): `WithOutputFrom()` now fails the build if the referenced executor is not present in the workflow graph
- **Self-loop detection** (warning): Logs a warning when an executor has an edge targeting itself, flagging potential infinite recursion
- **Dead-end detection** (info): Logs when executors have no outgoing edges, prompting review of intended terminal nodes
- Added tests for all new validation checks

### Other Changes
- Added README note about Go SDK evolving outside core MAF codebase
- Updated GitHub Actions lock files
- Added dependabot config


qmuntal (approved):
LGTM!

PR #138, opened 2026-05-06
Add tool approval middleware for HITL parity

gdams:
## Summary

Adds a `toolapproval` middleware package for .NET `ToolApprovalAgent` parity, closing the primary Human-in-the-Loop gap identified in the feature comparison.

### New: `agent/middleware/toolapproval`
- Implements the .NET `ToolApprovalAgent` pattern as idiomatic Go middleware
- **One-at-a-time surfacing**: When the inner agent returns multiple approval requests, only the first is surfaced to the caller; the rest are queued in session state
- **Standing rules ("don't ask again")**: Callers can respond with `AlwaysApproveToolResponse()` or `AlwaysApproveToolWithArgsResponse()` to create rules that auto-approve matching future tool calls for the session lifetime
- **Auto-approve loop**: When all approval requests match standing rules, the middleware automatically approves them and re-calls the inner agent without surfacing anything to the caller
- **Session persistence**: Rules, queued requests, and collected responses are persisted across turns via `agent.Session`

### Changes to `message` package
- Added `AlwaysApproveTool` and `AlwaysApproveToolWithArgs` fields to `FunctionApprovalResponseContent`
- Added `AlwaysApproveToolResponse()` and `AlwaysApproveToolWithArgsResponse()` convenience methods on `FunctionApprovalRequestContent`

### Tests
- Passthrough (no approval requests)
- First request surfacing with queuing
- Always-approve tool rule creation and auto-approval
- Queued requests surfaced one at a time
- AlwaysApproveToolWithArgs response creation

### DIFF
--- a/agent/middleware/toolapproval/toolapproval.go
+++ b/agent/middleware/toolapproval/toolapproval.go
@@ -288,6 +288,8 @@ func matchesRule(rules []Rule, req *message.ToolApprovalRequestContent) bool {
 	return false
 }
 
+// canonicalizeArguments normalizes JSON argument strings for stable equality
+// matching by unmarshaling and remarshaling them into canonical JSON.
 func canonicalizeArguments(arguments string) string {
 	var v any
 	if err := json.Unmarshal([]byte(arguments), &v); err != nil {
--- a/agent/middleware/toolapproval/toolapproval.go
+++ b/agent/middleware/toolapproval/toolapproval.go
@@ -24,8 +24,10 @@ import (
 	"github.com/microsoft/agent-framework-go/message"
 )
 
-const stateKey = "toolApprovalState"
-const rawArgumentsKey = "$raw"
+const (
+	stateKey        = "toolApprovalState"
+	rawArgumentsKey = "$raw"
+)
 
 // Rule is a standing approval rule. If Arguments is empty, all invocations
 // of the named tool are auto-approved. Otherwise only invocations with an
--- a/agent/middleware/toolapproval/toolapproval.go
+++ b/agent/middleware/toolapproval/toolapproval.go
@@ -1,0 +1,310 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// Package toolapproval provides middleware that manages human-in-the-loop
+// tool approval with support for "don't ask again" standing rules.
+//
+// When an inner agent returns tool calls that require approval, this
+// middleware surfaces approval requests one at a time to the caller. The
+// caller may approve or deny individual calls, and optionally indicate that
+// a tool (or a tool with specific arguments) should always be approved in
+// the future for the lifetime of the session.
+//
+// Standing approval rules and queued requests are persisted across turns
+// using the [agent.Session].
+package toolapproval
+
+import (
+	"context"
+	"encoding/json"
+	"iter"
+	"slices"
+
+	"github.com/microsoft/agent-framework-go/agent"
+	"github.com/microsoft/agent-framework-go/message"
+)
+
+const stateKey = "toolApprovalState"
+
+// Rule is a standing approval rule. If Arguments is empty, all invocations
+// of the named tool are auto-approved. Otherwise only invocations with an
+// exact argument match are auto-approved.
+type Rule struct {
+	ToolName  string `json:"toolName"`
+	Arguments string `json:"arguments,omitempty"`
+}
+
+// matches reports whether r auto-approves a call to toolName with the given
+// serialized arguments.
+func (r Rule) matches(toolName, arguments string) bool {
+	if r.ToolName != toolName {
+		return false
+	}
+	if r.Arguments == "" {
+		return true
+	}
+	return r.Arguments == arguments
+}
+
+// state is persisted in the session across turns.
+type state struct {
+	Rules              []Rule                                 `json:"rules,omitempty"`
+	CollectedResponses []*message.ToolApprovalResponseContent `json:"collectedResponses,omitempty"`
+	QueuedRequests     []*message.ToolApprovalRequestContent  `json:"queuedRequests,omitempty"`
+}
+
+func loadState(opts []agent.Option) state {
+	session, ok := agent.GetOption(opts, agent.WithSession)
+	if !ok {
+		return state{}
+	}
+	var s state
+	if found, _ := session.Get(stateKey, &s); found {
+		return s
+	}
+	return state{}
+}
+
+func saveState(opts []agent.Option, s state) {
+	session, ok := agent.GetOption(opts, agent.WithSession)
+	if !ok {
+		return
+	}
+	session.Set(stateKey, s)
+}
+
+// New creates a tool-approval middleware that wraps agent runs with
+// human-in-the-loop approval management.
+func New() agent.Middleware {
+	return agent.MiddlewareFunc(run)
+}
+
+func run(next agent.RunFunc, ctx context.Context, messages []*message.Message, opts ...agent.Option) iter.Seq2[*agent.ResponseUpdate, error] {
+	return func(yield func(*agent.ResponseUpdate, error) bool) {
+		st := loadState(opts)
+
+		// Step 1: Process inbound approval responses from the caller.
+		messages, st = prepareInbound(messages, st)
+
+		// Step 2: If we have queued requests from a previous turn, drain any
+		// that are now auto-approvable and surface the next one.
+		drainAutoApprovable(&st)
+		if len(st.QueuedRequests) > 0 {
+			next := st.QueuedRequests[0]
+			st.QueuedRequests = st.QueuedRequests[1:]
+			saveState(opts, st)
+			yield(&agent.ResponseUpdate{
+				Role:     message.RoleAssistant,
+				Contents: []message.Content{next},
+			}, nil)
+			return
+		}
+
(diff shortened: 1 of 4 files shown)


gdams:
> ## Cross-repo Parity Review — `toolapproval` middleware
> 
> This PR successfully fills the primary HITL gap and the overall logic (one-at-a-time surfacing, standing rules, queued drain, auto-approve loop) is semantically well-aligned with the upstream .NET `ToolApprovalAgent` in `dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/`. 
> 
> Two parity concerns worth resolving before merge:
> 
> ---
> 
> ### 1. Argument-matching strategy diverges from .NET (behavioral)
> 
> **Go** stores the raw JSON arguments string and compares it with `r.Arguments == arguments` (string equality on `json.Marshal` output).
> 
> **\.NET** stores arguments as `IDictionary<string, string>` (per-key serialized values) and matches with `ArgumentsMatch()`, which does key-by-key comparison — making matching **order-independent**.
> 
> Result: two semantically equivalent argument objects differing only in key ordering will match in .NET but **not** in Go. This is a behavioral parity gap when providers emit arguments with variable key ordering.
> 
> Upstream surface: [`ToolApprovalAgent.cs#ArgumentsMatch`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/ToolApprovalAgent.cs), [`ToolApprovalRule.cs`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/ToolApprovalRule.cs)
> 
> ---
> 
> ### 2. "Always approve" intent uses a different type shape (API design)
> 
> **\.NET** expresses "always approve" via a **distinct wrapper type** `AlwaysApproveToolApprovalResponseContent` (a separate `AIContent` subtype). The middleware detects it by type-switching, extracts the standing rule, and forwards only the unwrapped inner `ToolApprovalResponseContent` to the inner agent.
> 
> **Go** adds `AlwaysApproveTool bool` and `AlwaysApproveToolWithArgs bool` flags directly to the existing `FunctionApprovalResponseContent`. This means callers cannot distinguish the two cases by type, inner agents receive framework metadata they should not see, and the `AlwaysApproveToolWithArgs` name drifts from .NET's `AlwaysApproveToolWithArguments`.
> 
> Upstream surface: [`AlwaysApproveToolApprovalResponseContent.cs`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/AlwaysApproveToolApprovalResponseContent.cs), [`ToolApprovalRequestContentExtensions.cs`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/ToolApprovalRequestContentExtensions.cs)
> 
> ---
> 
> Both issues are inline-commented above. If the simpler Go-only design is intentional, please document the divergence in the package doc or a follow-up tracking issue.
> 
> 
> 
> 
> 
> > Generated by [Go API Consistency Review Agent](https://github.com/microsoft/agent-framework-go/actions/runs/25446650368/agentic_workflow) for issue #138 · ● 1.6M · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fagent-framework-go+%22gh-aw-workflow-call-id%3A+microsoft%2Fagent-framework-go%2Fgo-api-consistency-review%22&type=issues)
> 
> 
> 

@copilot push changes that address these divergences

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,344 @@
...
+
+const stateKey = "toolApprovalState"
+
+// Rule is a standing approval rule. If Arguments is empty, all invocations
+// of the named tool are auto-approved. Otherwise only invocations with an
+// exact canonical argument match are auto-approved, where canonical means JSON
+// normalized to a stable representation independent of object key ordering.
+type Rule struct {
+	ToolName  string `json:"toolName"`
+	Arguments string `json:"arguments,omitempty"`
```
qmuntal: .NET uses a `map[string]string` to represent arguments. Why don't we do the same?

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,344 @@
...
+	session, ok := agent.GetOption(opts, agent.WithSession)
+	if !ok {
+		return
+	}
+	session.Set(stateKey, s)
+}
+
+// New creates a tool-approval middleware that wraps agent runs with
+// human-in-the-loop approval management.
+func New() agent.Middleware {
```
qmuntal: Define a `type Config struct{}` and pass it as argument to `New`, as `func New(cfg Config) {}`. This will make this function future-proof. For example, we might want to pass json encoding options once `encoding/json/v2` is released. .NET already does something similar with optional parameters.

qmuntal (changes requested):
Looks good! Some minor comments,

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,344 @@
...
+				st.CollectedResponses = nil
+			}
+
+			var updates []*agent.ResponseUpdate
+			for update, err := range next(ctx, callMessages, opts...) {
+				if err != nil {
+					yield(nil, err)
+					return
+				}
+				updates = append(updates, update)
```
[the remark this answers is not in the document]
qmuntal: Good finding copilot 👍

message/content.go
```diff
@@ -558,6 +587,39 @@ func (t *FunctionApprovalResponseContent) MarshalJSON() ([]byte, error) {
...
+	InnerResponse *FunctionApprovalResponseContent `json:"innerResponse"`
+
+	// AlwaysApproveTool, when true, indicates a standing rule to auto-approve
+	// all future calls to the same tool regardless of arguments.
+	AlwaysApproveTool bool `json:"alwaysApproveTool,omitempty"`
+
+	// AlwaysApproveToolWithArguments, when true, indicates a standing rule to
+	// auto-approve future calls to the same tool only when the arguments
+	// match exactly.
+	AlwaysApproveToolWithArguments bool `json:"alwaysApproveToolWithArguments,omitempty"`
```
qmuntal: Don't use camel case, we just use the default Go's json encoding naming conventions as we don't need to be wire-compatible with .NET.

qmuntal (changes requested):

gdams:
@copilot linter failures

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,356 @@
...
+	if len(r.Arguments) == 0 {
+		return true
+	}
+	return argumentMapsEqual(r.Arguments, arguments)
+}
+
+// state is persisted in the session across turns.
+type state struct {
+	Rules              []Rule                                     `json:"rules,omitempty"`
+	CollectedResponses []*message.FunctionApprovalResponseContent `json:"collectedResponses,omitempty"`
```
qmuntal: the .NET sdk uses `ToolApprovalResponseContent` here. The Go equivalent is `message.ToolApprovalResponseContent`(which has just been added as part of  #152.

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,356 @@
...
+	QueuedRequests     []*message.FunctionApprovalRequestContent  `json:"queuedRequests,omitempty"`
```
qmuntal: the .NET sdk uses `ToolApprovalRequestContent` here. The Go equivalent is `message.ToolApprovalRequestContent`(which has just been added as part of  #152.

qmuntal (changes requested):

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,356 @@
...
+		return map[string]string{}
+	}
+	serialized := make(map[string]string, len(values))
+	for k, raw := range values {
+		serialized[k] = serializeArgumentValue(raw)
+	}
+	return serialized
+}
+
+func serializeArgumentValue(raw json.RawMessage) string {
```
qmuntal: Is this necessary? Converting a `map[string]json.RawMessage` into a `map[string]string` can be done by just casting json.RawMessage into string, if I'm not wrong.

```
serialized := make(map[string]string, len(values))
for k, v := range values {
  serialized[k] = string(v)
}
```

agent/middleware/toolapproval/toolapproval.go
```diff
@@ -0,0 +1,356 @@
...
+	args := serializeArguments(req.FunctionCall.Arguments)
+	for _, r := range rules {
+		if r.matches(req.FunctionCall.Name, args) {
+			return true
+		}
+	}
+	return false
+}
+
+func serializeArguments(arguments string) map[string]string {
```
qmuntal: .NET throws if the json encoding fails. This function returns `map[string]string{rawArgumentsKey: arguments}` instead. I would prefer to return an error that is properly propagated.

qmuntal (changes requested):

gdams:
@copilot resolve this merge conflict

qmuntal (approved):

PR #165, opened 2026-05-07
Add OpenTelemetry trace context propagation across workflow executors

gdams:
Fix the TODO in context.go: propagate OTel trace context across executor boundaries in workflow SendMessage/Bind. Adds otelutil helpers and tests.


qmuntal (changes requested):
I would like otel to be opt-in via some kind of abstraction so that people not wanting otel don't have otel on their dependency graph.

gdams:
@qmuntal updated PTAL

qmuntal (approved):
This PR goes in the right direction. Will do a follow-up PR with some improvements.

PR #202, opened 2026-05-20
[dotnet-port] Port A2A input-required task routing fix from .NET

qmuntal:
When an A2A task returns TaskStateInputRequired, follow-up messages should link to the waiting task via TaskId (direct link for input) rather than ReferenceTasks (context-only reference). This aligns with .NET PR microsoft/agent-framework#5743 which introduced the same fix.

- Track last task state in session state (a2aagent.taskState key)
- Pass task state to updateSessionContextID from all call sites, including yieldTask, sendMsg (Task/TaskStatusUpdateEvent events), and the GetTask continuation path
- Use userMsg.TaskId when last task state is InputRequired (with a known task ID); fall back to ReferenceTasks otherwise
- Add TestRunWithInputRequiredTask_UsesTaskId and TestRunWithCompletedTask_UsesReferenceTasks to cover both paths

### DIFF
--- a/agent/provider/a2aagent/a2a.go
+++ b/agent/provider/a2aagent/a2a.go
@@ -75,7 +75,7 @@ func (a *a2aagent) run(ctx context.Context, messages []*message.Message, options
 				yield(nil, err)
 				return
 			}
-			if err := updateSessionContextID(session, task.ContextID, string(task.ID)); err != nil {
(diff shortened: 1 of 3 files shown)


qmuntal:
@copilot fix feedback

gdams (approved):

PR #203, opened 2026-05-20
Align workflow APIs with .NET parity

qmuntal:
Summary:
- Add ProtocolBuilder-based protocol declaration and update call sites to use singular variadic send/yield methods.
- Align message routing with .NET behavior for exact and assignable interface handlers.
- Make Workflow topology builder-owned/private and update runtime/test access through read-only methods.

Tests:
- go test ./...


qmuntal:
@copilot fix golang-lint

gdams (approved):

PR #220, opened 2026-05-26
Improve workflow executor bindings and examples

qmuntal:
## Summary
- Move workflow executor behavior onto `Executor`, including lifecycle hooks, `Extend`, `SetCrossRunShareable`, implementation IDs, and struct marker attributes for declared send/yield types.
- Add `NewExecutor`, `BindNewExecutorFunc`, and `RequestPort.Bind` so functions, structs with `Handle`, existing executors/bindings, and request ports can be adapted into workflow executors.
- Update builder validation, routing, telemetry, message workflow helpers, and workflow hosting adapters around implementation IDs and action handlers that should not auto-send or auto-yield return values.
- Refresh workflow examples and docs, including new start-here, agents, checkpoint human-in-the-loop, map-reduce, workflow-as-agent, observability, and Azure OpenAI helper examples.
- Expand workflow, inproc, message workflow, hosting, provider, and example tests for the new executor/binding behavior.

## Testing
- Focused VS Code test run for changed workflow, message workflow, workflow hosting, and provider test files: 172 passed.


qmuntal:
@copilot fix golangci-lint

gdams (approved):

PR #228, opened 2026-05-28
Group OpenTelemetry gomod updates in Dependabot

qmuntal:
## Summary
- add a Dependabot gomod group named opentelemetry
- include OpenTelemetry root, sdk, trace, and stdouttrace modules
- keep existing update cadence and PR limits unchanged

## Why
This keeps related OpenTelemetry dependency bumps in a single PR for easier review and testing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,13 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+  groups:
+    opentelemetry:
+      patterns:
+        - "go.opentelemetry.io/otel"
+        - "go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,6 +8,13 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+  groups:
+    opentelemetry:
+      patterns:
+        - "go.opentelemetry.io/otel"
+        - "go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
```
gdams:
```suggestion
        - "go.opentelemetry.io/otel/*"
```

gdams (changes requested):
We can use wildcards here

gdams (approved):

PR #350, opened 2026-06-24
[dotnet-port] Align external response port correlation with .NET

michelle-clayton-work:
This pull request strengthens the validation logic for handling external responses in the workflow engine. It ensures that responses are only accepted if they match the expected request and port, and improves test coverage to verify this behavior. The main changes are grouped below.

**Validation and Error Handling Improvements:**

* Replaced the previous `CompleteRequest` method with a new `completeResponse` method in `runnerContext`, which validates that an incoming response matches both the expected request ID and port ID before marking it as complete. If the response does not match, an error is returned. (`workflow/inproc/context.go`)
* Updated `AddExternalResponse` to use the new `completeResponse` method, so that mismatched responses are correctly rejected with an error. (`workflow/inproc/context.go`)

**Testing Enhancements:**

* Improved the test `TestRequestPortBind_RejectsResponseForOtherPort` to explicitly check for error events when a response is received for a mismatched port, and to verify that a legitimate response after the error is handled correctly without producing additional errors. (`workflow/inproc/binding_test.go`)

### DIFF
--- a/workflow/inproc/binding_test.go
+++ b/workflow/inproc/binding_test.go
@@ -1009,9 +1009,30 @@ func TestRequestPortBind_RejectsResponseForOtherPort(t *testing.T) {
 		Data:      workflow.AnyPortableValue(int(7)),
 	}
 	if _, err := run.Resume(ctx, resp); err != nil {
-		t.Fatalf("Resume: %v", err)
+		t.Fatalf("Resume with mismatched response port: %v", err)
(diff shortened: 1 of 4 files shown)


michelle-clayton-work:
@copilot fix PR feedback.

michelle-clayton-work:
@copilot new branch + push

PR #402, opened 2026-06-30
Consolidate provider integration packages

qmuntal:
## Goal

This PR makes provider integrations easier to discover by giving each integration family one obvious home. Provider-specific agent adapters, hosting adapters, tests, and examples now live under the corresponding `provider/<name>provider` package instead of being split across `agent/provider`, `agent/hosting`, `agent/opentelemetry`, and `agent/memory`.

The main intent is package navigation: users looking for OpenAI, A2A, AGUI, Gemini, Anthropic, Copilot, Foundry memory, or OpenTelemetry support should be able to start in `provider/` and find both the provider-facing agent constructors and the related hosting/integration surfaces in the same area.

## What changed

Most provider packages moved from `agent/provider/*agent` to `provider/*provider`, and examples/docs were updated to use the new paths. A2A and AGUI hosting were also folded directly into `provider/a2aprovider` and `provider/aguiprovider`, so hosting APIs sit beside the provider APIs rather than in separate `*hosting` subpackages. During that merge, overlapping private helpers were consolidated: A2A hosting reuses the parent provider content conversion helpers, and AGUI hosting reuses the provider tool-result serialization helper.

Workflow agent integrations were consolidated under `workflow/agentworkflow`. That package now owns both directions of the agent/workflow bridge: wrapping workflows as agents and hosting agents inside workflows, including the sequential, concurrent, and group-chat workflow builders. This keeps workflow-specific bridging near the workflow package while still removing the old split between `agent/hosting/workflowhosting` and `agent/provider/workflowprovider`.

The PR also updates README and feature-comparison documentation so the public package map matches the new organization.

## Validation

Ran the full Go test suite after the final package moves:

```text
go test ./...
```

Also ran `git diff --check --cached` before committing.


qmuntal:
@copilot fix golangci-lint failure

gdams (approved):

PR #405, opened 2026-06-30
Align context and history providers with .NET

qmuntal:
## Summary
- align context and history provider lifecycles with the .NET Agent Framework patterns (`Invoking`/`Invoked` plus shared invocation context objects)
- make context and history providers interface-based with additive default implementations and .NET-aligned filter names
- update default and in-memory history behavior, including `InMemoryHistoryProviderConfig` with state key, initializer, and filter hooks
- migrate built-in providers, examples, and tests to the revised provider APIs

## Tests
- go test ./...
- git diff --cached --check

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -57,7 +57,7 @@ type Config struct {
 
 	// HistoryProvider injects and persists conversation history around each agent run.
 	// When nil, New uses a default in-memory history provider for local sessions.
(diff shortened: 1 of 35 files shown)


gdams (approved):

qmuntal:
All cross parity findings are not relevant and missed the point.

PR #421, opened 2026-07-02
Align data URI and URI content behavior with MEAI

qmuntal:
## Summary
- default data URIs without an explicit media type to `text/plain;charset=US-ASCII`
- normalize non-base64 data URI payloads through URL decoding before storing base64 data
- add `NewURIContent(uri, mediaType)` with optional-by-empty-string media type inference from URI file extension

## Upstream MEAI parity
- dotnet/extensions#7247
- dotnet/extensions#7398

## Validation
- `go test ./message`

### DIFF
--- a/message/content.go
+++ b/message/content.go
@@ -533,6 +533,18 @@ type URIContent struct {
 	URI       string
 }
 
+func NewURIContent(uri string, mediaType string) (*URIContent, error) {
+	if err := validateURIContentURI(uri); err != nil {
+		return nil, err
+	}
+	if mediaType == "" {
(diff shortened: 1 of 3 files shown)


gdams:
> Non-base64 data URI decoding — ⚠️ Minor divergence (more RFC-correct in Go)

@qmuntal any concerns about this?

qmuntal:
> @qmuntal any concerns about this?

That's fine. We are a bit more compliant with how data URI are decoded. I tried to align with .NET, but that meant reimplement most of the complex URI decoding code.

gdams (approved):

PR #437, opened 2026-07-06
[TEST PR -- do not review] Dev/mclayton/verify aw locks ci test

michelle-clayton-work:


gdams:
@michelle-clayton-work this wasn't really what I had in mind. I was thinking of telling the dependabot config to ignore all `.github/workflows/*.lock.yml` files using a glob. I believe that dependabot has a way to exclude paths. It might make sense for us to write a tool in go-infra that we use to apply these changes to a whole bunch of repos.

PR #464, opened 2026-07-09
cmd/prfromissue: open and link PRs for gh-aw fallback issues

michelle-clayton-work:
The porting workflows run with a token that cannot open pull requests, so gh-aw's create-pull-request safe output pushes the branch and files a tracking issue with a compare link instead, leaving the eventual PR unlinked from the issue.

prfromissue scans open issues by title prefix, parses the pushed branch/base/title out of the fallback issue body, opens the pull request with a 'Closes #<issue>' body, and comments the PR link back on the issue. When a PR already exists for the branch it links the two instead of opening a duplicate. It shells out to the gh CLI (local 'gh auth login' or GH_TOKEN), and supports -dry-run to preview.

Local test:
image


michelle-clayton-work:
@copilot address PR feedback

gdams (approved):

cmd/prfromissue/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+### Flags
+
+| Flag       | Default | Description                                                        |
+| ---------- | ------- | ------------------------------------------------------------------ |
+| `-prefix`  | —       | Required. Only act on issues whose title starts with this prefix.  |
+| `-repo`    | ""      | `owner/repo`; defaults to the repository in the current directory. |
+| `-limit`   | 100     | Maximum number of open issues to scan.                             |
+| `-draft`   | true    | Open the pull request as a draft.                                  |
```
gdams: I'm not sure that opening as draft should be the default, perhaps switch it around?

cmd/prfromissue/main.go
```diff
@@ -0,0 +1,100 @@
...
+//
+// Authentication uses the gh CLI: `gh auth login` locally, or the
+// GH_TOKEN / GITHUB_TOKEN environment variable in CI.
+//
+// Usage:
+//
+//	go run ./cmd/prfromissue -prefix "[dotnet-port-api]"
+//	go run ./cmd/prfromissue -prefix "[dotnet-port-fixes]" -repo microsoft/agent-framework-go
+//	go run ./cmd/prfromissue -prefix "[dotnet-port-api]" -dry-run
+//	go run ./cmd/prfromissue -prefix "[dotnet-port-api]" -draft=false
```
dagood: Is the user really expected to always include a prefix, and map that correctly to whether it should be a draft or not, etc.? Could these expectations be put in the tool so all the user needs to do is a single call to `go run ./cmd/prfromissue`?

cmd/prfromissue/main.go
```diff
@@ -0,0 +1,100 @@
...
+// Command prfromissue opens pull requests for gh-aw fallback tracking issues and
+// links the two together.
+//
+// The agent-framework porting workflows run with a token that cannot open pull
+// requests, so gh-aw's create-pull-request safe output pushes the branch and
+// files a tracking issue containing a compare link instead. This tool scans open
+// issues whose title starts with a given prefix (e.g. "[dotnet-port-api]"),
+// opens the pull request from the pushed branch, sets its body to "Closes
+// #<issue>", and comments the PR link back on the issue. When a PR already exists
+// for the branch it links the two instead of opening a duplicate.
```
dagood: Copilot doesn't see any filtering based on username or anything else that a random user couldn't submit as an issue. I don't know if that's necessarily a vulnerability in itself (what this input can make the tool do, ultimately), but seems like it could lead to one if a few things get put together.

An issue author filter would be a simple approach, although it's not ideal that it all happens as "app/github-actions". Some other workflow with a vuln could potentially file an issue as that user and include some malicious instructions.

cmd/prfromissue/main.go
```diff
@@ -0,0 +1,100 @@
...
+	draft  bool
+	dryRun bool
+}
+
+func parseOptions(args []string) (options, bool) {
+	fs := flag.NewFlagSet("prfromissue", flag.ContinueOnError)
+	fs.SetOutput(os.Stderr)
+	var opts options
+	fs.StringVar(&opts.prefix, "prefix", "", "required: only act on open issues whose title starts with this prefix (e.g. \"[dotnet-port-api]\")")
+	fs.StringVar(&opts.repo, "repo", "", "target repository as owner/repo (defaults to the repository in the current directory)")
```
dagood: A git clone directory doesn't contain a repository, although it could have some remotes that point at GitHub repositories. Maybe I'm missing how this is intended to be used, but I'm not sure it makes sense to do a default based on context vs. just pointing at this repository.

Personally, when I write a tool like this, I want to make it not care about the directory, so I can `go install` it and run it from a basic terminal that isn't in any particular directory.

dagood (commented):
A few high-level comments about the tool.

cmd/prfromissue/main.go
michelle-clayton-work: Good point. Updating so that the default look at all issues (no prefix) and keeping the prefix as an optional flag. I think this is better than defaulting to the three prefixes to allow for easily adding more workflows and wit hthe added issue-author that already basically limits the issues anyway

cmd/prfromissue/main.go
michelle-clayton-work: Fair. Adding an issue-author allowlist flag where the default is app/github-actions and the user can make the flag empty to remove the allowlist

cmd/prfromissue/main.go
michelle-clayton-work: Agreed. updated so that the default is agentframework

PR #470, opened 2026-07-11
Fix duplicate and corrupted streamed Anthropic tool calls

PratikDhanave:
## Summary

Fixes two related bugs in `provider/anthropicprovider`'s streaming path that break streamed tool calls, most visibly with parallel tool use:

1. **Duplicate tool call emission.** Every `content_block_stop` event re-emitted *all* accumulated function calls (the `functions` map was walked in full and never pruned), instead of only the block that stopped. With two or more `tool_use` blocks in one streamed response — Anthropic's normal shape for parallel tool calls — the first call was yielded again on every later block stop. Downstream consumers such as the `toolautocall` middleware accumulate `FunctionCallContent` without CallID dedup, so the same tool call was **executed more than once**.

2. **Corrupted argument JSON.** A streamed `tool_use` `content_block_start` event carries only an input placeholder (`"input": {}`); the real arguments arrive via `input_json_delta` events. `buildBlock` seeded `Arguments` with that placeholder and the deltas were appended after it, so every streamed tool call produced invalid arguments like `{}{"city":"Paris"}`.

## Fix

In the streaming loop:
- `content_block_stop` now emits only `functions[event.Index]` and deletes it from the map, so each function call is yielded exactly once, when its block completes.
- The start-event input placeholder is set aside so `input_json_delta` fragments accumulate from scratch; if a block produces no deltas, the placeholder is restored on stop, preserving prior behavior for calls whose input arrives on the start event.

The non-streaming path is unchanged (its single end-of-response map walk was already correct).

No public API changes.

## Tests

- Added `TestStreaming_ParallelToolCalls_NoDuplicates`: replays an SSE stream with two parallel `tool_use` blocks through an `httptest.Server` (matching the existing test style) and asserts each `CallID` is yielded exactly once with correctly assembled arguments. The test fails on `main` with `FunctionCallContent for toolu_01 yielded 2 times, want 1` and `toolu_01 arguments = "{}{\"city\":\"Paris\"}"`.
- `go test -race -shuffle=on ./...` passes.
- `gofmt`/`gofumpt` clean, `go vet` clean.

### DIFF
--- a/provider/anthropicprovider/agent.go
+++ b/provider/anthropicprovider/agent.go
@@ -127,6 +127,7 @@ func (a *client) run(ctx context.Context, messages []*message.Message, options .
 		var modelID string
 		var usage message.UsageDetails
 		functions := make(map[int]*message.FunctionCallContent)
+		inputPlaceholders := make(map[int]string)
 
 		for stream.Next() {
 			event := stream.Current()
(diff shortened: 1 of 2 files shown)


qmuntal:
Thanks for identifying and fixing the duplicate and corrupted streamed Anthropic tool-call behavior. This is superseded by #490, which addresses both issues using the Anthropic Go SDK's built-in message accumulator and adds coverage for fragmented, multiple, and interleaved tool-call streams.

PR #473, opened 2026-07-12
Fix panic on chat completion response with no choices

PratikDhanave:
## Summary

The non-streaming Chat Completions path in `provider/openaiprovider/chat.go` indexes `resp.Choices[0]` unconditionally after a successful API call. Some OpenAI-compatible services return HTTP 200 with an **empty `choices` array** — most notably Azure OpenAI when the prompt is blocked by a content filter, where the response carries `prompt_filter_results` and no choices. Any such response panics before reaching the caller:

```
panic: runtime error: index out of range [0] with length 0
  openaiprovider.(*chatClient).run(...)
      provider/openaiprovider/chat.go:117
```

The streaming path in the same file already guards every `chunk.Choices` access with `len(chunk.Choices) > 0`; the non-streaming path was missing the equivalent guard.

## Fix

Return an error (`chat completion response contained no choices`) through the response stream instead of panicking, mirroring the existing error paths in `run`.

No public API changes.

## Tests

- Added `TestChatEmptyChoices_NonStreaming`: replays an Azure-content-filter-shaped response (200, `"choices": []`, `prompt_filter_results`) through the existing `newTestServer` harness and asserts `Collect()` returns an error. On `main` it fails with the panic above; with the fix it passes.
- `go test -race -shuffle=on ./...` passes.
- `gofmt`/`gofumpt` clean on the changed files, `go vet` clean.

### DIFF
--- a/provider/openaiprovider/chat.go
+++ b/provider/openaiprovider/chat.go
@@ -6,6 +6,7 @@ import (
 	"cmp"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"iter"
 	"net/http"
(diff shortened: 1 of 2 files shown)


PratikDhanave:

qmuntal:
To align both modes with .NET behavior, choice-less responses should be tolerated rather than treated as errors. Please guard access to Choices[0] and return an update containing any response metadata and usage.

PratikDhanave:
Updated to tolerate choice-less responses instead of erroring: the non-streaming path now guards `Choices[0]` and emits an update with the response metadata + usage, matching the streaming path and .NET behavior. `TestChatEmptyChoices_NonStreaming` now asserts the response is tolerated (no error) and that usage is surfaced. Thanks for the steer!

qmuntal (approved):

PR #491, opened 2026-07-14
Fix data race in agentmode provider session state

PratikDhanave:
## Problem

The `agentmode` provider's `mode_set`/`mode_get` tools and its exported `GetMode`/`SetMode` methods read and write the agent session state (`agent.Session.Get`/`Set`, a plain `map`) with no synchronization. In addition, the tool closures shared a single `*state` pointer captured from `provide()` and mutated it directly.

`toolautocall` invokes multiple tool calls from a single model response on **separate goroutines** when `AllowConcurrentInvocations` is enabled (see `autocall.go`, the `len(funcCalls) > 1 && f.allowConcurrentInvocations` branch). So two concurrent `mode_set`/`mode_get` calls in one turn race on both the session map and the shared `*state`.

Running the added test under `-race` on `main` reports races at `agentmode.go` (`st.CurrentMode` read/write) and `session.go` (`Session.Set` map write).

The sibling `todo` provider already guards the identical session access with per-session locking — `agentmode` was the inconsistent one.

## Fix

Mirror the `todo` provider's approach:

- Add a `sync.Map` of per-session mutexes keyed by `session.ServiceID()` (falling back to a shared `nullSessionLock` when no session is present), via a `getSessionLock` helper.
- Take the lock around every session-state access: both tool closures, `provide()`, `GetMode`, and `SetMode`.
- Reload state inside the lock in each tool closure instead of mutating a `*state` captured across invocations, eliminating the shared-pointer race.

No public API changes.

## Test

Adds `TestConcurrentToolInvocations_NoDataRace`, which primes the provider, then invokes `mode_set`/`mode_get` concurrently on separate goroutines sharing one session — mirroring the `AllowConcurrentInvocations` path. It fails under `-race` on `main` and passes with the fix.

## Verification

- `go test -race -shuffle=on ./agent/...` — passing
- `go vet`, `gofmt`, `gofumpt` — clean
- `go build ./...` — clean


agent/harness/agentmode/agentmode_test.go
```diff
@@ -48,6 +49,57 @@ func collectInstructions(opts []agent.Option) string {
...
+			defer wg.Done()
+			_, _ = setTool.Call(context.Background(), `{"Arg0":"`+mode+`"}`)
+		}(mode)
+		go func() {
+			defer wg.Done()
+			_, _ = getTool.Call(context.Background(), "")
+		}()
+	}
+	wg.Wait()
+}
```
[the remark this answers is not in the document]
PratikDhanave: Good point — updated so each goroutine records its Call error in a distinct slot (no added race on the slice) and the test asserts them all after the wait. A functional regression now fails deterministically, independent of the race detector.

qmuntal:
Thanks for addressing the race. Reloading state under the lock for each tool invocation is the right approach.

One important change is still needed: please key the lock registry by session object identity instead of `Session.ServiceID()`.

Using the service ID has several problems:

- Locally managed sessions commonly have an empty service ID, causing unrelated sessions to share the `"_default"` lock.
- Distinct sessions can have the same service ID and therefore unexpectedly share a lock.
- The service ID is mutable, so the same session could resolve to different locks during its lifetime, undermining the synchronization guarantee.
- Entries in the `sync.Map` are never removed and will accumulate for the provider’s lifetime.
- A nil session should use `nullSessionLock` rather than the `"_default"` entry.

Since the module targets Go 1.25, consider an identity-based weak-key registry using `weak.Pointer[agent.Session]`. Dead entries should be removed either opportunistically during lookup or with `runtime.AddCleanup`, because weak pointers alone do not remove map entries.

Please also add tests confirming that:

- the same `*agent.Session` always receives the same lock;
- different sessions with empty service IDs receive different locks;
- different sessions with the same service ID receive different locks;
- a nil session uses the shared fallback lock;
- stale registry entries do not accumulate indefinitely.

Stable lock identity is essential to the race fix: a mutable key can allow one session to use multiple locks, while a non-unique key creates unnecessary cross-session contention.

PratikDhanave:
Thanks — good call, keying by service ID was fragile. Reworked to key the registry by session object identity using `weak.Pointer[agent.Session]`:

- **Stable & unique key**: object identity never changes and is unique per session, so a session always maps to the same lock and distinct sessions never share one — regardless of service ID (empty, shared, or later mutated).
- **nil/absent session** now uses `nullSessionLock` instead of a shared `"_default"` entry.
- **No unbounded growth**: since weak keys neither keep sessions alive nor remove map entries, `runtime.AddCleanup` deletes each entry once its session is collected. The cleanup captures only the weak key, never the session.

Added internal tests covering all five cases you listed: same session → same lock; distinct sessions with empty service IDs → distinct locks; distinct sessions with the same service ID → distinct locks; nil/absent session → fallback lock; and stale entries draining after GC (the last polls a bounded deadline since `AddCleanup` runs asynchronously).

Also merged latest `main`. `go test -race -shuffle=on ./agent/...` is green.

Side note: the sibling `todo` provider uses the same `ServiceID`-keyed pattern this replaces, so it likely has the same latent issue — happy to file a follow-up for it if useful.

agent/harness/agentmode/getsessionlock_internal_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Can we move these tests to agentmode_test.go? I prefer to always do black box testing when possible.

qmuntal (changes requested):

agent/harness/agentmode/getsessionlock_internal_test.go
PratikDhanave: Agreed on preferring black-box. I removed the white-box `getSessionLock` tests: the lock-identity properties they checked (same session → same lock, distinct sessions → distinct locks, GC drain) are an implementation detail with no clean black-box observation — distinct vs. shared locks are both race-free, so nothing external can distinguish them. The behavior that actually matters — concurrency safety under concurrent tool invocation — is already covered black-box by `TestConcurrentToolInvocations_NoDataRace` in `agentmode_test.go`.

Happy to restore them as an explicit white-box file if you'd rather keep direct identity coverage — just let me know which you prefer.

qmuntal (approved):

qmuntal (approved):

PR #495, opened 2026-07-14
otelprovider: record token usage on the agent span

### LINKED ISSUE #493: otelprovider: agent spans carry no token usage or model, so there is no cost visibility (the data is already in the stream)
## Summary

`otelprovider`'s span carries the agent's **identity** but not its **cost**. There is currently no way to see token usage or the model on a trace, even though both are already available at the point where the span is written.

The result is a stopwatch with good labels: you can see that an agent ran and for how long, but not what it spent.

## What the span emits today

Running an agent with `otelprovider.NewMiddleware(...)` attached:

```
span: BrainAgent   18.14s
  gen_ai.operation.name    = invoke_agent
  gen_ai.provider.name     = anthropic
  gen_ai.agent.id          = e7f04186-...
  gen_ai.agent.name        = BrainAgent
  gen_ai.agent.description =
```

That is the whole attribute set.

## What is missing, and why it is easy

The OpenTelemetry GenAI semantic conventions define all of these, **and the constants are already in the `semconv` package this repo imports**:

- `gen_ai.usage.input_tokens`  (`semconv.GenAIUsageInputTokens`)
- `gen_ai.usage.output_tokens` (`semconv.GenAIUsageOutputTokens`)
- `gen_ai.request.model`       (`semconv.GenAIRequestModel`)

And the data is already flowing past the middleware. `anthropicprovider` emits usage into the response stream (`provider/anthropicprovider/agent.go:109`):

```go
contents = append(contents, &message.UsageContent{
    Details: toUsageDetails(resp.Usage),
})
```

`message.UsageDetails` already carries `InputTokenCount`, `OutputTokenCount`, `TotalTokenCount`, `CachedInputTokenCount`, `ReasoningTokenCount`.

So the middleware, which already iterates the stream, can accumulate `*message.UsageContent` and set the attributes on the span it already opened. I grepped the repo: `input_tokens` currently appears **only inside serialization helpers**, never as a span attribute.

## Why this is the attribute set people actually want

Token count is cost, and cost is the main reason to trace an agent at all. Concretely, from a real run of a 5-agent workflow once I added these attributes locally:

```
workflow:ticket-to-brief   303s
  researcher   66s  turns=6  tool_calls=14  in=197,057  out=3,195
  grapher      45s  turns=2  tool_calls=5   in=111,940  out=2,512
  author       27s  turns=2  tool_calls=5   in= 10,493  out=1,957
  critic       84s  turns=4  tool_calls=12  in= 75,790  out=5,515
  publisher    35s  turns=2  tool_calls=2   in= 15,692  out=3,414
  ────────────────────────────────────────────────────────────
  TOTAL  410,972 in / 16,593 out
```

The researcher costing ~2x the author is not visible without this, and it is the single most useful thing the trace can tell you.

## Suggested addition

In the middleware's stream loop, accumulate `message.UsageDetails` and set on span end:

- `gen_ai.request.model` (from `AgentConfig.Model`; the middleware would need it passed via `MiddlewareConfig`, or the provider could set it)
- `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` / `gen_ai.usage.total_tokens`
- optionally `gen_ai.usage.cached_input_tokens` and `gen_ai.usage.reasoning_tokens` when non-zero

**Deliberately NOT suggesting cost.** Prices go stale and are deployment-specific; a framework asserting a stale price is worse than one asserting none. Exposing usage is the right primitive — let people price it themselves.

I have this working locally and am happy to open a PR.

lifeofzero:
Fixes #493.

## The problem

`otelprovider`'s span carries the agent's **identity** but not its **cost**. The whole attribute set today is:

```
gen_ai.operation.name    = invoke_agent
gen_ai.provider.name     = anthropic
gen_ai.agent.id / name / description
```

That tells you an agent ran and for how long. It does not tell you what it spent — and token count is cost, which is the main reason to trace an agent at all.

## Why it is a small change

The numbers were **already passing through this middleware**. Providers emit `message.UsageContent` into the response stream (`provider/anthropicprovider/agent.go`), and this middleware already iterates that stream on its way to the caller. It just never recorded them.

No new dependencies. No behaviour change. Just reading what goes past.

## What it adds

Per the OpenTelemetry GenAI semantic conventions:

| attribute | |
|---|---|
| `gen_ai.usage.input_tokens` | always |
| `gen_ai.usage.output_tokens` | always |
| `gen_ai.usage.total_tokens` | always |
| `gen_ai.usage.cached_input_tokens` | only when non-zero |
| `gen_ai.usage.reasoning_tokens` | only when non-zero |

## Three deliberate choices, in case you disagree with any

1. **Summed across the run, not last-write-wins.** An agent that makes several LLM round-trips emits one `UsageContent` per round-trip. Reporting only the final one understates a multi-turn agent's real spend — which is exactly the agent whose cost you most want to see.

2. **Optional counters are omitted when zero, not written as `0`.** "This provider does not report cached tokens" and "this provider reports zero cached tokens" are different facts. An attribute that is always present and always zero trains people to ignore it.

3. **No cost or pricing.** Prices go stale and are deployment-specific; a framework asserting a stale price is worse than one asserting none. Usage is the right primitive — callers can price it themselves, and that is what I do downstream.

## What it looks like in practice

From a 5-agent workflow, once these attributes exist:

```
workflow:ticket-to-brief   303s
  researcher   66s  in=197,057  out=3,195
  grapher      45s  in=111,940  out=2,512
  author       27s  in= 10,493  out=1,957
  critic       84s  in= 75,790  out=5,515
  publisher    35s  in= 15,692  out=3,414
  ──────────────────────────────────────
  TOTAL             410,972 in / 16,593 out
```

The researcher costing ~20x the author is not visible without this, and it is the single most actionable thing the trace can say.

## Tests

- `TestOtel_Run_RecordsTokenUsage` — asserts the **sum** across two turns (100+200 in, 10+20 out), and that `reasoning_tokens` is **absent** when the provider reports none.
- `TestOtel_Run_NoUsageAttributesWhenProviderReportsNone` — a run with no usage emits no usage attributes at all.

Both fail without the change (verified by reverting the accumulation: `span is missing gen_ai.usage.input_tokens`). Full suite passes; `go vet` and `gofmt` clean.

### DIFF
--- a/provider/otelprovider/otel.go
+++ b/provider/otelprovider/otel.go
@@ -48,11 +48,14 @@ const (
 	// stream -- so the numbers pass through here on their way to the caller and were
 	// simply never recorded. Token count is cost, and cost is the main reason to trace
 	// an agent at all.
-	attrKeyUsageInputTokens       = "gen_ai.usage.input_tokens"
-	attrKeyUsageOutputTokens      = "gen_ai.usage.output_tokens"
-	attrKeyUsageTotalTokens       = "gen_ai.usage.total_tokens"
-	attrKeyUsageCachedInputTokens = "gen_ai.usage.cached_input_tokens"
-	attrKeyUsageReasoningTokens   = "gen_ai.usage.reasoning_tokens"
+	//
+	// Names are the exact ids from the GenAI semantic-conventions registry. The registry
+	// namespaces the cache/reasoning counters (they are not `<x>_tokens`) and defines no
+	// total; do not infer these from the input/output shape.
+	attrKeyUsageInputTokens     = "gen_ai.usage.input_tokens"
+	attrKeyUsageOutputTokens    = "gen_ai.usage.output_tokens"
+	attrKeyUsageCacheReadTokens = "gen_ai.usage.cache_read.input_tokens"
+	attrKeyUsageReasoningTokens = "gen_ai.usage.reasoning.output_tokens"
 )
 
 type mw struct {
@@ -83,13 +86,9 @@ func (m *mw) Run(next agent.RunFunc, ctx context.Context, messages []*message.Me
 				span.RecordError(err, trace.WithTimestamp(time.Now()))
 				span.SetStatus(codes.Error, err.Error())
 			}
-			if update != nil {
-				for _, content := range update.Contents {
-					if u, ok := content.(*message.UsageContent); ok {
-						usage.Add(u.Details)
-					}
-				}
-			}
+			// update.Usage() sums this update's UsageContent (nil-safe), so accumulate
+			// its total into the run rather than iterating Contents by hand.
+			usage.Add(update.Usage())
 			if !yield(update, err) {
 				// Set what we have before bailing. A caller that stops reading early
 				// still ran (and paid for) the tokens counted so far, and a span that
@@ -116,13 +115,15 @@ func setUsage(span trace.Span, usage message.UsageDetails) {
 		return
 	}
 
+	// input + output are the only usage-token attributes the registry defines (no
+	// total); cache_read is a subset of input and reasoning a subset of output, so
+	// consumers sum rather than reading a provider-side total.
 	attrs := []attribute.KeyValue{
 		attribute.Int64(attrKeyUsageInputTokens, usage.InputTokenCount),
 		attribute.Int64(attrKeyUsageOutputTokens, usage.OutputTokenCount),
-		attribute.Int64(attrKeyUsageTotalTokens, usage.TotalTokenCount),
 	}
 	if usage.CachedInputTokenCount > 0 {
-		attrs = append(attrs, attribute.Int64(attrKeyUsageCachedInputTokens, usage.CachedInputTokenCount))
+		attrs = append(attrs, attribute.Int64(attrKeyUsageCacheReadTokens, usage.CachedInputTokenCount))
 	}
 	if usage.ReasoningTokenCount > 0 {
 		attrs = append(attrs, attribute.Int64(attrKeyUsageReasoningTokens, usage.ReasoningTokenCount))
--- a/provider/otelprovider/otel.go
+++ b/provider/otelprovider/otel.go
@@ -108,7 +108,11 @@ func (m *mw) Run(next agent.RunFunc, ctx context.Context, messages []*message.Me
 // tokens should be distinguishable from one that reports none, and an attribute that
 // is always present but always zero trains people to ignore it.
 func setUsage(span trace.Span, usage message.UsageDetails) {
-	if usage.InputTokenCount == 0 && usage.OutputTokenCount == 0 && usage.TotalTokenCount == 0 {
+	// "Did we observe ANY usage?" -- and that must consider every counter, not just the
+	// three required ones. Guarding on input/output/total alone would silently drop the
+	// whole attribute set for a provider that reported only cached or reasoning tokens,
+	// which is the exact silent-drop this function exists to avoid.
+	if !hasUsage(usage) {
 		return
 	}
 
@@ -124,4 +128,15 @@ func setUsage(span trace.Span, usage message.UsageDetails) {
 		attrs = append(attrs, attribute.Int64(attrKeyUsageReasoningTokens, usage.ReasoningTokenCount))
 	}
 	span.SetAttributes(attrs...)
+}
+
+// hasUsage reports whether any counter was populated. UsageDetails carries a map
+// (AdditionalCounts) so it is not comparable with ==; the fields are checked directly.
+func hasUsage(u message.UsageDetails) bool {
+	return u.InputTokenCount != 0 ||
+		u.OutputTokenCount != 0 ||
+		u.TotalTokenCount != 0 ||
+		u.CachedInputTokenCount != 0 ||
+		u.ReasoningTokenCount != 0 ||
+		len(u.AdditionalCounts) > 0
 }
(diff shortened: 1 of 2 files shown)


lifeofzero:

lifeofzero:
Both addressed in `baf9d65`. Thanks — the first one was a real bug and a slightly embarrassing one.

**1. `setUsage` dropping optional-only counters.** You are right, and the failure is ironic: a function whose entire purpose is to not silently drop usage was silently dropping usage. Guarding on `input/output/total` answers the wrong question. The question is "did we observe **any** usage?", which has to consider every counter.

Fixed with an explicit `hasUsage()`. (`UsageDetails` carries a map — `AdditionalCounts` — so it is not comparable with `==` and a zero-value check will not compile.)

Added a regression test for exactly the case you described — a provider reporting **only** `CachedInputTokenCount` — and verified it catches the old behaviour:

```
--- FAIL: TestOtel_Run_RecordsUsageWhenOnlyOptionalCountersReported
    otel_test.go:547: cached_input_tokens was dropped when it was the only counter reported
```

**2. Shadowed loop variable.** Renamed to `attr` in both tests. You are right that it is a trap for future edits.

Full suite still green, `go vet` and `gofmt` clean.

provider/otelprovider/otel.go
```diff
@@ -38,6 +38,19 @@ const (
...
+	// Token usage, per the OpenTelemetry GenAI semantic conventions.
+	//
+	// Providers already emit message.UsageContent into the response stream (see
+	// provider/anthropicprovider/agent.go), and this middleware already iterates that
+	// stream -- so the numbers pass through here on their way to the caller and were
+	// simply never recorded. Token count is cost, and cost is the main reason to trace
+	// an agent at all.
+	attrKeyUsageInputTokens       = "gen_ai.usage.input_tokens"
+	attrKeyUsageOutputTokens      = "gen_ai.usage.output_tokens"
+	attrKeyUsageTotalTokens       = "gen_ai.usage.total_tokens"
```
qmuntal: This is not a standard otel attribute, and I'm not aware of any related attribute. Remove.

provider/otelprovider/otel.go
```diff
@@ -38,6 +38,19 @@ const (
...
+	attrKeyUsageCachedInputTokens = "gen_ai.usage.cached_input_tokens"
```
qmuntal: The standard otel attribute name is [gen_ai.usage.cache_read.input_tokens](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-usage-cache-read-input-tokens).

qmuntal: The standard otel attribute name is [gen_ai.usage.cache_read.input_tokens](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-usage-cache-read-input-tokens).

provider/otelprovider/otel.go
```diff
@@ -38,6 +38,19 @@ const (
...
+	attrKeyUsageReasoningTokens   = "gen_ai.usage.reasoning_tokens"
```
qmuntal: The standard otel attribute name is [gen_ai.usage.reasoning.output_tokens](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-usage-reasoning-output-tokens).

provider/otelprovider/otel.go
```diff
@@ -56,14 +69,70 @@ func (m *mw) Run(next agent.RunFunc, ctx context.Context, messages []*message.Me
...
+		// the agent rather than the cost of its final call.
+		var usage message.UsageDetails
+
 		for update, err := range next(ctx, messages, options...) {
 			if err != nil {
 				span.RecordError(err, trace.WithTimestamp(time.Now()))
 				span.SetStatus(codes.Error, err.Error())
 			}
+			if update != nil {
+				for _, content := range update.Contents {
```
qmuntal: Use update.Usage() instead of a custom loop.

qmuntal (changes requested):

lifeofzero:
Thanks for the review, and apologies for the churn. All four are addressed in `b95295c`, and this time I verified each name against the GenAI semantic-conventions registry rather than inferring it from the `input_tokens`/`output_tokens` shape:

- Removed `gen_ai.usage.total_tokens`. The registry defines no total attribute; per the spec, `cache_read` is a subset of `input_tokens` and reasoning a subset of `output_tokens`, so consumers sum. Added a test asserting we do not emit a total.
- `cached_input_tokens` -> `gen_ai.usage.cache_read.input_tokens`.
- `reasoning_tokens` -> `gen_ai.usage.reasoning.output_tokens`.
- Replaced the hand-rolled `UsageContent` loop with `update.Usage()` (nil-safe, so the `if update != nil` guard is gone too).

Tests, `go vet`, and `gofmt` are green.

One nice-to-have I looked into but deliberately left out of this PR: OTel Go ships generated `semconv` constants for all of these (`semconv/v1.41.0`, already reachable via the `otel v1.44.0` this module pins) as `GenAIUsageInputTokens`, `GenAIUsageOutputTokens`, `GenAIUsageCacheReadInputTokens`, `GenAIUsageReasoningOutputTokens`. Using those instead of string literals would make a wrong or non-existent name a compile error (there is no `GenAIUsageTotal` to reference), binding the middleware to the spec at build time. I kept it out here because the repo does not use `semconv` anywhere yet, so adopting it is a convention decision beyond this rename fix and one I would not want to make unilaterally. If you would like it, I am happy to do it as a follow-up, either scoped to the usage attributes or converting all the `gen_ai.*` attributes in this file for consistency. Your call.

dagood (dismissed):

lifeofzero:
Rebased onto `main` to resolve the conflict with #446 (kept its `error.type` / `otelx` additions; my usage recording sits alongside them). gofmt, go vet, and the otelprovider tests are green. Ready for re-review.

qmuntal (approved):
Thanks!

PR #496, opened 2026-07-14
anthropicprovider: add opt-in prompt caching (cache_control)

lifeofzero:
## The gap

`anthropicprovider` **reads cached-token counts back** — `CacheReadInputTokens` and `CacheCreationInputTokens` at `agent.go:188` — but **never sets `cache_control`**. There is not a single occurrence of `cache_control` or `ephemeral` in the provider.

So it faithfully reports cache hits it is **structurally incapable of producing**. Prompt caching is simply unreachable from this provider today.

## Why it matters here more than most places

A tool-using agent re-sends its **entire context on every turn**: the tool definitions, the system prompt, and every tool result accumulated so far. That repeated prefix *is* the bill.

Measured on a real 5-agent workflow against an MCP toolserver:

```
410,972 input tokens   vs   16,593 output tokens
→ input is 83% of the cost, and most of it is the same bytes over and over
→ cached_input_tokens = 0 on every span
```

One agent alone made **21 tool calls whose results were 5–20KB each**, and carried all of them forward through 8 turns.

## What this adds

`AgentConfig.PromptCaching` (default **off**), placing two breakpoints:

1. **End of the system prompt.** Anthropic's cacheable prefix is ordered `tools → system → messages`, so this single breakpoint covers the tool **definitions** *and* the instructions — the part that is byte-identical on every turn.
2. **End of the last message.** Caches the conversation so far, including the tool **results**, which is where a tool-using agent's tokens actually live.

Uses 2 of Anthropic's 4 allowed breakpoints, leaving room for finer control via the existing `MessageNewParams` option.

## Two deliberate decisions, in case you'd rather go the other way

**Opt-in, not on by default.** A cache *write* costs more than a plain input token, so caching is a net **loss** for a short single-turn agent with a small prompt — precisely the agent whose author is least likely to be measuring. Defaulting it on would quietly raise the bill for the cheapest use case in order to lower it for the most expensive one. I think that trade should be a decision, not a surprise.

**Every content-block variant is handled, not just text.** The trailing block on a tool-using turn is usually a `tool_result`, not a `text` block. Handling only text would leave the expensive half of the context uncached *while the config still reported caching as on* — caching that claims success and saves nothing, which is worse than no caching at all.

## Tests

Three, asserting the **wire format** (marshalled JSON) rather than SDK struct internals — `cache_control` is a contract with the API, and that is the thing that has to be right.

All three are mutation-verified:

| mutation | result |
|---|---|
| ignore the `PromptCaching` flag | `TestPromptCaching_OffByDefault` fails |
| drop the `OfToolResult` case (text-only breakpoint) | `TestPromptCaching_MarksTrailingToolResultBlock` fails with *"the expensive half of the context would go uncached"* |

Full suite green, `go vet` and `gofmt` clean.

Related: #493 / #495 (token usage on spans) — that PR is what made this measurable in the first place.

### DIFF
--- a/provider/anthropicprovider/agent.go
+++ b/provider/anthropicprovider/agent.go
@@ -490,9 +490,14 @@ func buildMessageParam(msg *message.Message) (anthropic.MessageParam, error) {
 	var content []anthropic.ContentBlockParamUnion
 
 	for _, c := range msg.Contents {
+		var (
+			block anthropic.ContentBlockParamUnion
+			built bool
+		)
 		switch c := c.(type) {
 		case *message.TextContent:
-			content = append(content, anthropic.NewTextBlock(c.Text))
+			block = anthropic.NewTextBlock(c.Text)
+			built = true
 		case *message.FunctionCallContent:
 			// Parse the JSON string arguments into a map for Anthropic SDK
 			var args map[string]any
@@ -507,7 +512,8 @@ func buildMessageParam(msg *message.Message) (anthropic.MessageParam, error) {
 				// for a tool call with empty or absent arguments.
 				args = map[string]any{}
 			}
-			content = append(content, anthropic.NewToolUseBlock(c.CallID, args, c.Name))
+			block = anthropic.NewToolUseBlock(c.CallID, args, c.Name)
+			built = true
 		case *message.FunctionResultContent:
 			resStr := ""
 			switch {
@@ -535,16 +541,28 @@ func buildMessageParam(msg *message.Message) (anthropic.MessageParam, error) {
 					}
 				}
 			}
-			content = append(content, anthropic.NewToolResultBlock(c.CallID, resStr, c.Error != nil))
+			block = anthropic.NewToolResultBlock(c.CallID, resStr, c.Error != nil)
+			built = true
 		case *message.DataContent:
 			if c.TopLevelMediaType() == "image" {
 				mediaType := c.MediaType
 				if mediaType == "" {
 					mediaType = "image/jpeg"
 				}
-				content = append(content, anthropic.NewImageBlockBase64(mediaType, c.Data))
+				block = anthropic.NewImageBlockBase64(mediaType, c.Data)
+				built = true
 			}
 		}
+		if !built {
+			continue
+		}
+		// Opt-in cache_control: if the source content was marked via WithCacheControl,
+		// carry that breakpoint onto the block just built for it. A content with no marker
+		// is left untouched, so caching never turns itself on silently. This runs for every
+		// block type -- text, image, document, tool-use, tool-result -- because the marker
+		// lives on the message.Content, not on any particular Anthropic wire type.
+		applyContentCacheControl(c, &block)
+		content = append(content, block)
 	}
 
 	switch msg.Role {
--- a/provider/anthropicprovider/caching.go
+++ b/provider/anthropicprovider/caching.go
@@ -1,0 +1,167 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+package anthropicprovider
+
+import (
+	"reflect"
+
+	"github.com/anthropics/anthropic-sdk-go"
+	"github.com/microsoft/agent-framework-go/message"
+)
+
+// Ephemeral5m and Ephemeral1h are the cache lifetimes Anthropic supports for a
+// cache_control breakpoint. They are the anthropic-sdk-go values, surfaced here as
+// named constants so callers do not have to reach for the SDK's verbose identifiers.
+//
+// Their type is the SDK's own anthropic.CacheControlEphemeralTTL, deliberately: this
+// package already hands raw SDK types across its public surface (see MessageNewParams),
+// so a provider-specific caching primitive stays consistent by doing the same rather
+// than wrapping the SDK enum in a parallel type the caller would have to convert.
+const (
+	// Ephemeral5m caches for 5 minutes. This is also the default when no TTL is set.
+	Ephemeral5m = anthropic.CacheControlEphemeralTTLTTL5m
+	// Ephemeral1h caches for 1 hour.
+	Ephemeral1h = anthropic.CacheControlEphemeralTTLTTL1h
+)
+
+// CacheControl describes an Anthropic cache_control breakpoint on a single content block.
+type CacheControl struct {
+	// TTL is the cache lifetime. The zero value ("") means the Anthropic default of
+	// 5-minute ephemeral caching; Ephemeral1h selects the 1-hour tier.
+	TTL anthropic.CacheControlEphemeralTTL
+}
+
+// CacheControlOption configures the CacheControl attached by WithCacheControl.
+type CacheControlOption func(*CacheControl)
+
+// WithTTL sets the cache lifetime for the breakpoint. Passing the zero value leaves
+// Anthropic's default (5 minutes) in effect.
+func WithTTL(ttl anthropic.CacheControlEphemeralTTL) CacheControlOption {
+	return func(cc *CacheControl) { cc.TTL = ttl }
+}
+
+// cacheControlKey is the package-private key under which a cache_control marker is stored
+// in a content's AdditionalProperties bag. Unexported and package-qualified so it cannot
+// collide with a key a caller sets for their own purposes.
+const cacheControlKey = "anthropicprovider.cacheControl"
+
+// WithCacheControl attaches an Anthropic cache_control breakpoint to a single
+// message.Content, in place, and returns it for chaining. Caching is expressed on the
+// individual content block, not toggled agent-wide, so the caller decides exactly which
+// blocks become cache breakpoints.
+//
+// The marker is stored in the content's ContentHeader.AdditionalProperties under a
+// package-private key and read back in buildMessageParam, which transfers it onto whatever
+// Anthropic block the content becomes: text, image, document, tool-use, or tool-result. A
+// content whose block type has no cache_control field on the wire (the thinking and
+// redacted_thinking variants) is a clean no-op rather than an error.
+//
+// Opt-in, deliberately. A cache WRITE costs more than a plain input token, so caching is a
+// net loss for a short single-turn prompt; nothing here turns it on unless the caller asks.
+//
+// Breakpoint budget: Anthropic permits at most 4 cache_control breakpoints per request
+// (the tool definitions, the system prompt, and message blocks all draw from the same
+// budget). The caller is responsible for staying within it; this function does not dedupe
+// or cap, and a request with too many breakpoints is rejected by the API. Higher-level
+// automatic placement is intentionally left to a future change.
+//
+// For full control over the raw request, including breakpoints on tools or the system
+// prompt, use MessageNewParams as the low-level escape hatch.
+func WithCacheControl(c message.Content, opts ...CacheControlOption) message.Content {
+	var cc CacheControl
+	for _, opt := range opts {
+		opt(&cc)
+	}
+	markCacheControl(c, cc)
+	return c
+}
+
+// markCacheControl writes the cache_control marker into a content's AdditionalProperties.
+//
+// Every message.Content implementation embeds message.ContentHeader by value, so each one
+// already carries an AdditionalProperties bag for provider metadata. ContentHeader.Header()
+// only hands back a COPY, though, so the field has to be reached through the content's own
+// pointer to mutate the original. Reflection does that generically: it finds the promoted
+// AdditionalProperties field on any content type, present and future, without a per-type
+// switch that would silently miss a type the day it is added.
+//
+// TODO: this reflection dance can be dropped once message.Content exposes its ContentHeader
+// by pointer (Header() returns a copy today). With a pointer accessor, markCacheControl can
+// set AdditionalProperties on the original header directly, with no reflection.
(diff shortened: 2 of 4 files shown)


provider/anthropicprovider/agent.go
```diff
@@ -427,9 +452,73 @@ func (a *client) buildMessageParams(messages []*message.Message, opts []agent.Op
...
+	// 2. End of the last message (covers the conversation so far).
+	//
+	// The breakpoint must go on a content block, so it is placed on the last block of
+	// the last message. Setting it on the message itself is not a thing the API models.
+	if n := len(params.Messages); n > 0 {
+		blocks := params.Messages[n-1].Content
+		if len(blocks) > 0 {
+			setCacheControl(&blocks[len(blocks)-1])
+		}
+	}
```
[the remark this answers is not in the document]
lifeofzero: Good catch, and fixed in 492b515 — but it turned out to be the small half of the bug.

You're right that a trailing `thinking` block swallows the breakpoint: `thinking` has no `cache_control` field on the wire, so caching would be *on* and not one message token cached. Silent, which is exactly the failure this PR exists to prevent. (It isn't reachable today — `buildMessageParam` has no thinking case — but it's latent the moment one lands.)

Chasing it surfaced the bigger problem. `ContentBlockParamUnion` has **17 variants**; the switch covered five. So ten *cacheable* variants were being silently skipped — including every server-tool result (`web_search_tool_result`, `code_execution_tool_result`, `tool_search_tool_result`, …), which are the **largest** blocks in a modern agent's context. Precisely the ones you most want cached, quietly uncached.

Rather than extend the list to 17 and have it rot on the 18th, `setCacheControl` now reflects over the union and makes the SDK's own types the source of truth: **a block is cacheable exactly when its param struct carries a `CacheControl` field.** `thinking` and `redacted_thinking` are the only two that don't. It returns a bool, and `applyPromptCaching` walks backward to the last block that accepts a breakpoint — an earlier breakpoint still caches nearly the entire prefix, so degraded beats silent. Cost is one reflect walk over a 17-field struct per request, next to a network call to an LLM.

Three tests, each verified to fail against the previous code:
- `SkipsTrailingThinkingBlock` — your case.
- `FallsBackToEarlierMessage` — an all-thinking final message still gets a breakpoint.
- `CoversEveryCacheableVariant` — drives itself off the union's reflected shape, so a new SDK variant is covered the day it lands. Reverting to a hand-written switch fails it loudly, naming all ten missed variants.

qmuntal:
Thanks for identifying this gap and for the thorough implementation and tests.

We would like this to follow the approach used by the .NET Anthropic SDK. Rather than adding an agent-level `PromptCaching` flag and automatically selecting cache breakpoints, .NET makes caching explicit on individual content blocks through `WithCacheControl(...)`. The provider then carries that metadata into the corresponding Anthropic request block.

Could you rework this PR accordingly?

- Remove `AgentConfig.PromptCaching` and the automatic breakpoint placement.
- Add an Anthropic-specific mechanism for attaching cache-control metadata to individual `message.Content` values.
- Preserve that metadata when converting supported content into Anthropic text, image, document, tool-use, and tool-result blocks.
- Support the available Anthropic cache-control settings, including TTL.
- Keep `MessageNewParams` as the low-level escape hatch.
- Retain wire-format tests verifying supported content types and confirming that unmarked content remains unchanged.

References:

- https://github.com/anthropics/anthropic-sdk-csharp/blob/main/src/Anthropic/AIContentCacheExtensions.cs
- https://github.com/anthropics/anthropic-sdk-csharp/commit/f007149c471aa40a1ed7ab34e611cc749213f5ed

This keeps prompt caching explicit and opt-in, gives callers control over cost and breakpoint placement, and aligns the Go and .NET provider APIs. The work already done here—particularly around tool-result and unsupported thinking blocks—should remain useful for the per-content implementation.

lifeofzero:
Thanks, this makes sense and I'm happy to rework it this way. Explicit per-content cache control is the better fit: it matches how Anthropic models `cache_control` (a per-block field), keeps placement and cost decisions with the caller, and lines the Go API up with .NET.

Plan for the reworked PR:

- Drop `AgentConfig.PromptCaching` and the automatic breakpoint placement.
- Add an Anthropic-specific `WithCacheControl(...)` that attaches cache-control metadata to an individual `message.Content` (via its `AdditionalProperties`, mirroring the .NET `AIContent` extension).
- Carry that metadata through `buildMessageParam` onto the corresponding text, image, document, tool-use, and tool-result blocks.
- Support the ephemeral cache-control settings including TTL (`5m` / `1h`), which the Go SDK's `CacheControlEphemeralParam` already exposes.
- Keep `MessageNewParams` as the low-level escape hatch.
- Keep the wire-format tests: one per supported content type, plus a check that unmarked content is emitted unchanged.

As you noted, the tool-result handling and the unsupported-block cases (thinking / redacted_thinking carry no `cache_control` field) carry straight over to the per-content path, so that part stays.

One optional question, and I'm fine either way: the common agentic-loop pattern is always the same shape (cache the system prompt, the last turn, and large stable tool results). Would a small opt-in helper layered on top of `WithCacheControl` for that case be welcome, or would you rather keep the provider surface minimal and leave all placement to callers? Happy to leave it out and keep this PR to just the primitive.

PratikDhanave:
This is a genuinely excellent PR — I read it closely and wanted to flag what stands out, plus one question.

**What's especially strong:**

- The opt-in-by-default reasoning, stated right in the field doc: a cache *write* costs more than a plain input token, so defaulting it on would quietly tax the cheap single-turn agent to subsidize the expensive one. Rare to see that tradeoff made explicit.
- The reflection-over-switch call in `setCacheControl`, plus `TestSetCacheControl_CoversEveryCacheableVariant` driving off the union's own reflected shape (and guarding the guard with the 10+/2 counts). I checked `ContentBlockParamUnion` on SDK v1.57 — 17 `OfXxx` variant pointers plus an embedded by-value `paramUnion` that your `Kind() != Pointer` check correctly skips — so the "exactly one non-nil pointer = the variant" invariant holds, and the canary locks it for the 18th variant. A hand-written switch really would silently drop every server-tool result.
- "Degraded beats silent" — walking back past a trailing `thinking`/`redacted_thinking` block, with regression tests for both the same-message and earlier-message fallbacks. Right philosophy for a cost feature, where a silent no-op is the worst outcome.

**One question (not a blocker):** when `PromptCaching` is on *and* a caller also sets `cache_control` via the `MessageNewParams` option you mention, `applyPromptCaching` adds 2 breakpoints unconditionally. Anthropic caps `cache_control` at 4 breakpoints and 400s beyond that — so a caller who's already placed 3 would trip it. Worth a sentence in the `PromptCaching` doc ("if you set cache_control manually, keep the combined total ≤ 4"), or a cheap guard? It's the one interaction I couldn't find handled.

_(Not a maintainer — just another contributor who learned something reading this. Deferring to the team on the merge.)_

qmuntal:
> Would a small opt-in helper layered on top of WithCacheControl for that case be welcome, or would you rather keep the provider surface minimal and leave all placement to callers? Happy to leave it out and keep this PR to just the primitive.

Let's leave higher level constructs for a future PR.

lifeofzero:
Reworked per @qmuntal's guidance to make caching explicit and per-content, matching the .NET `WithCacheControl` model. Force-pushed as a single squashed commit, so the earlier inline threads are now against the old approach.

**What changed**

- Removed `AgentConfig.PromptCaching` and the automatic breakpoint placement.
- Added `WithCacheControl(content, opts...)` in the anthropicprovider package. It attaches cache-control metadata to an individual `message.Content` by storing it in that content's `ContentHeader.AdditionalProperties` under a package-private key, mirroring how .NET's `AIContent` carries provider metadata.
- `buildMessageParam` reads the marker back and transfers `cache_control` onto whatever Anthropic block the content becomes: text, image, document, tool-use, tool-result. A block whose wire type has no `cache_control` field (thinking / redacted_thinking) is a clean no-op rather than an error.
- TTL via `WithTTL(Ephemeral5m | Ephemeral1h)`. The TTL type is the SDK's own `anthropic.CacheControlEphemeralTTL` rather than a wrapper, to stay consistent with the package already handing raw SDK types across its surface (`MessageNewParams`).
- `MessageNewParams` remains the low-level escape hatch for breakpoints on tools or the system prompt.

The reflection-based block marking that a few of you looked at is retained (generalized to carry the TTL), and the variant-coverage canary test is kept and extended.

**On the 4-breakpoint question** (thanks @PratikDhanave for catching it) — with automatic placement gone, the old "2 unconditional + your manual ones" interaction no longer exists. The cap is still real, so I documented it on `WithCacheControl`: Anthropic allows at most 4 `cache_control` breakpoints per request and the caller is responsible for staying within that budget; this primitive does not dedupe or cap, and higher-level automatic placement is intentionally left to a future PR per @qmuntal. No runtime validation, matching .NET.

`go build ./...`, `go test ./provider/anthropicprovider/...`, and `go vet` are green.

provider/anthropicprovider/caching.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// cacheControlKey is the package-private key under which a cache_control marker is stored
+// in a content's AdditionalProperties bag. Unexported and package-qualified so it cannot
+// collide with a key a caller sets for their own purposes.
+const cacheControlKey = "anthropicprovider.cacheControl"
+
+// WithCacheControl attaches an Anthropic cache_control breakpoint to a single
+// message.Content, in place, and returns it for chaining. It is the Go analogue of the
+// .NET Anthropic SDK's per-AIContent cache-control extension: caching is expressed on the
```
qmuntal: Don't mention .NET in the public Go docs.

provider/anthropicprovider/caching.go
```diff
@@ -0,0 +1,165 @@
...
+		opt(&cc)
+	}
+	markCacheControl(c, cc)
+	return c
+}
+
+// markCacheControl writes the cache_control marker into a content's AdditionalProperties.
+//
+// Every message.Content implementation embeds message.ContentHeader by value, so each one
+// already carries an AdditionalProperties bag -- the same hook .NET's AIContent uses for
```
qmuntal: Don't mention .NET in the public Go docs.

provider/anthropicprovider/caching_internal_test.go
```diff
@@ -0,0 +1,289 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal, line 1: Use black box testing and move all tests to provider\anthropicprovider\agent_test.go

provider/anthropicprovider/caching.go
```diff
@@ -0,0 +1,165 @@
...
+// provider metadata. ContentHeader.Header() only hands back a COPY, though, so the field
+// has to be reached through the content's own pointer to mutate the original. Reflection
+// does that generically: it finds the promoted AdditionalProperties field on any content
+// type, present and future, without a per-type switch that would silently miss a type the
+// day it is added.
+func markCacheControl(c message.Content, cc CacheControl) {
+	v := reflect.ValueOf(c)
```
qmuntal, line 92: Hmm, unfortunate that `message.Content` doesn't return a pointer to `ContentHeader`. Will fix that in a follow up PR. Add a TODO saying that all this reflection dance can be dropped when that happens.

qmuntal (changes requested):
Thank!

karianna:
@lifeofzero will need a rebase

lifeofzero:
Thanks for the review @qmuntal. Addressed, and rebased onto `main` (the `agent.go` tool_use conflict resolved to keep the new nil-args guard alongside the per-content `cache_control` build path):

- Dropped the two `.NET` references from the public Go docs.
- Added a TODO on `markCacheControl`: the reflection can go away once `message.Content` exposes its `ContentHeader` by pointer (`Header()` returns a copy today).
- Moved the behavioral wire-format tests to `agent_test.go` as black-box (package `anthropicprovider_test`) — they now go through the public `agent.Run` path and assert on the captured outbound request.

On "move all tests": I kept exactly one test internal — `TestSetCacheControl_CoversEveryCacheableVariant`. It reflects over the SDK's `ContentBlockParamUnion` to prove every variant that carries a `cache_control` field receives a breakpoint (and that `thinking`/`redacted_thinking` report false). The public agent path only produces a handful of the SDK's 17 block variants, so a black-box test can't exercise the other twelve — which is the whole point of that guard: catching a future SDK variant, or a hand-written switch that silently skips the server-tool results. I left a comment in the file explaining why it stays package-internal. Happy to drop it if you'd rather have nothing internal, but it's the safety net the reflective approach rests on — let me know your preference.

lifeofzero:
@qmuntal i'm offgrid for the next ten days. so if this doesn't meet the bar feel free to close it. unfortunately i can't do more.

gdams:
@lifeofzero sorry about the delay in getting back to you here. Would you mind resolving the conflicts and fixing the linter and then we can get this merged

lifeofzero:
@gdams done, just waiting on the workflows to be approved.

qmuntal:
There are still some unresolved review comments.

provider/anthropicprovider/caching.go
qmuntal: Done in #868.

lifeofzero:
Sorry, @qmuntal I don't have the cycles to complete this right now.

qmuntal:
No problem. Closing, thanks anyway!

PR #497, opened 2026-07-14
Clarify SearchDepth scope in fsskills documentation

PratikDhanave:
## Summary

Following review, this PR is repurposed from a behavior change to a documentation clarification.

`SearchDepth` controls only resource and script discovery *within* an already-discovered skill directory. Skill-directory discovery (locating `SKILL.md` directories) is bounded independently and intentionally, matching the .NET SDK. That scope was not explicit in the option's documentation, which is what led to the original misread.

## Change

- Document the intended scope on `SourceOptions.SearchDepth` and on `searchForSkills`.
- Add `TestFileSource_SearchDepth_DoesNotAffectSkillDirectoryDiscovery`, which asserts that a large `SearchDepth` does **not** widen skill-directory discovery — locking in the intended contract.

No behavior change; the functional code is identical to `main`.

## Verification

- `go test -race -shuffle=on ./agent/skills/...` — passing
- `go vet`, `gofmt`, `gofumpt` — clean

### DIFF
--- a/agent/skills/fsskills/source.go
+++ b/agent/skills/fsskills/source.go
@@ -62,6 +62,12 @@ type SourceOptions struct {
 	// files within each skill directory. A value of 1 searches only the skill
 	// root; a value of 2 (the default) also searches one level of subdirectories.
 	// Values less than 1 are treated as the default.
+	//
+	// SearchDepth applies only to resource and script discovery within a skill
+	// directory. It does not affect the discovery of skill directories
+	// themselves (those containing a SKILL.md), which is bounded independently.
+	// This mirrors the .NET SDK, where the two concerns are configured
+	// separately.
 	SearchDepth int
 
 	// ResourceFilter is an optional predicate applied to each candidate resource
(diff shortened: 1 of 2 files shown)


qmuntal:
Thanks for identifying and documenting this behavior.

After comparing it with the .NET SDK, the current Go behavior is intentional. Skill-directory discovery is bounded independently at two levels, while `SearchDepth` controls resource and script discovery within an already discovered skill directory.

Making `SearchDepth` control skill-directory discovery would broaden its existing semantics and cause the Go SDK to diverge from .NET. For cross-SDK consistency, we won’t merge this change.

Could you repurpose this PR to update the Go documentation so the intended scope of `SearchDepth` is explicit?

PratikDhanave:
Thanks for the clarification and the .NET comparison — that makes sense. I've repurposed this PR to a documentation-only change: it now documents that `SearchDepth` scopes to within-skill resource/script discovery only, and that skill-directory discovery is bounded independently. I also added a test that locks in the intended contract (a large `SearchDepth` does not widen directory discovery). No behavior change.

qmuntal (approved):

PR #504, opened 2026-07-16
Guard inproc runner checkpoint state against concurrent access

PratikDhanave:
## Summary

The `inproc` runner writes its `checkpoints` and `lastCheckpointInfo` fields from the **background run-loop goroutine** during supersteps (`runner.go`, the checkpoint step) and on restore, while consumers read them concurrently through the public `StreamingRun.Checkpoints()` / `StreamingRun.LastCheckpoint()` accessors (`run.go`). These fields had **no synchronization** — the struct uses `atomic.Bool` only for `needsRepublish` — so `slices.Clone(r.checkpoints)` in `Checkpoints()` races the `append(r.checkpoints, …)` in the checkpoint step.

Polling checkpoint progress from another goroutine during a live streaming run is a supported usage, so this is reachable through the ordinary public API whenever checkpointing is enabled.

Race detector output (on `main`):
```
WARNING: DATA RACE
  Write ... runner.go:472  (checkpoint step: r.checkpoints = append(...))
  Read  ... runner.go:289  (Checkpoints(): slices.Clone(r.checkpoints))
```

## Fix

Add a `checkpointMu sync.Mutex` guarding `checkpoints` and `lastCheckpointInfo`, and lock all four access sites: `Checkpoints()`, the checkpoint step's parent read and post-commit writes, and the restore path.

The mutex is **not** held across `checkpointMgr.Commit(ctx, …)` (a blocking I/O call): the parent info is read under the lock, released, committed, then the result is published under the lock again. Critical sections stay limited to the field accesses so the public accessors remain cheap.

## Public API

No exported symbols change; the fix is internal to the runner.

## Tests

Adds `TestStreamingRun_ConcurrentCheckpointAccess_NoDataRace`, which polls `Checkpoints()`/`LastCheckpoint()` from a goroutine while a streaming checkpointed run advances through initial run, response, and restore. It **fails under `-race` before this change** (the write/read race above) and **passes after**.

`go test -race -shuffle=on ./workflow/inproc` passes; `go build ./...` and `go vet ./workflow/...` clean; gofumpt clean.


workflow/inproc/checkpoint_race_test.go
```diff
@@ -0,0 +1,73 @@
...
+		for {
+			select {
+			case <-stop:
+				return
+			default:
+				_ = run.Checkpoints()
+				_, _ = run.LastCheckpoint()
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
PratikDhanave: This is intentional for a race-repro test: the tight poll maximizes the overlap window with the background checkpoint writes so the race detector fires reliably, and it's bounded by the short test lifetime (it stops as soon as the run halts). A ticker would space the reads out and could miss the write window, making the repro flaky — which defeats the test's purpose. Happy to add a small pacing delay if you'd prefer, but I'd lean on keeping detection deterministic.

workflow/inproc/checkpoint_race_test.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Move these tests to checkpoint_test.go.

qmuntal (changes requested):

workflow/inproc/checkpoint_race_test.go
PratikDhanave: Done — moved into `checkpoint_test.go`.

qmuntal (approved):

PR #507, opened 2026-07-16
Only track conditionless edges in the conditionless-edge dedup set

PratikDhanave:
## Summary

`Builder.AddDirectEdge` appends **every** connection to `wb.conditionlessConnections` (builder.go), including *conditional* edges, but the duplicate check fires only when `condition == nil` and treats any entry in that list as a pre-existing **conditionless** edge. So a conditional edge added first on a `source→target` pair poisons the dedup set: a subsequent legitimate conditionless edge on the same pair is wrongly rejected.

Concrete trigger:
```go
b.AddDirectEdge(a, b, false, func(any) bool { return true }) // conditional
b.AddEdge(a, b)                                              // conditionless
// => wb.err = "an edge from 'a' to 'b' already exists without a condition"
```
even though no conditionless edge exists. Via `AddChain` (which calls `AddDirectEdge` with `idempotent=true`), the same poisoning instead **silently drops** the chain edge (the dedup returns early without adding it).

## Fix

Guard the append with `condition == nil`, so only conditionless edges populate the conditionless-edge dedup set — matching the field's name and sole purpose (it is read only by that dedup check).

## Public API

No exported symbols change.

## Tests

- `TestBuilder_ConditionalEdgeDoesNotBlockConditionlessEdge` — conditional then conditionless edge on the same pair now builds without error (fails before this change with the misleading duplicate error).
- `TestBuilder_ConditionalEdgeDoesNotDropIdempotentConditionlessEdge` — the idempotent path doesn't silently drop the conditionless edge.

`go test ./workflow` passes; gofmt clean.


workflow/builder_conditional_edge_test.go
```diff
@@ -0,0 +1,43 @@
...
+	wf, err := workflow.NewBuilder(start).
+		AddDirectEdge(start, target, false, func(any) bool { return true }).
+		AddDirectEdge(start, target, true, nil). // idempotent conditionless edge
+		Build()
+	if err != nil {
+		t.Fatalf("unexpected build error: %v", err)
+	}
+	if wf == nil {
+		t.Fatal("expected a workflow")
+	}
```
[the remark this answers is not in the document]
PratikDhanave: Good catch — strengthened the test to inspect `wf.Edges()["start"]` and assert both a conditional and a conditionless edge are present, so it now fails if the idempotent conditionless edge is silently dropped.

workflow/builder_conditional_edge_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Move tests to builder_test.go

qmuntal (changes requested):

workflow/builder_conditional_edge_test.go
PratikDhanave: Done — moved both tests into builder_test.go and removed the separate file.

qmuntal (approved):

PR #509, opened 2026-07-16
Report Gemini streaming usage once, not per chunk

PratikDhanave:
## Summary

The `geminiprovider` streaming path appended a `*message.UsageContent` for **every** stream chunk whose `UsageMetadata` was non-nil (provider/geminiprovider/agent.go). Gemini reports `usageMetadata` **cumulatively** across chunks — the final chunk's totals are authoritative — and the downstream `Response.Usage()` aggregation sums every `UsageContent` via `UsageDetails.Add`. So a response that carries usage on more than one chunk **over-counts** tokens:

- chunk 1 running total 12, chunk 2 running total 15 → reported **27** instead of **15**.

The non-streaming path emits usage once and was already correct. (The provider's existing streaming test placed usage only on the final chunk, so this path was untested.)

## Fix

Track the latest `UsageMetadata` during the stream and emit a single `UsageContent` once the stream ends, so aggregation reflects the authoritative final total rather than the sum of running totals.

## Public API

No exported symbols change.

## Tests

Adds `TestUsageContent_Streaming_CumulativeUsageNotSummed`: two chunks carrying cumulative usage (totals 12 then 15). `resp.Usage().TotalTokenCount` is **27 before this change** and **15 after** (with `InputTokenCount` 10, `OutputTokenCount` 5).

`go test ./provider/geminiprovider` passes; gofmt clean.


provider/geminiprovider/agent.go
```diff
@@ -159,6 +162,13 @@ func (a *client) run(ctx context.Context, messages []*message.Message, options .
 				return
 			}
 		}
+		if latestUsage != nil {
+			yield(&agent.ResponseUpdate{
+				Contents:  []message.Content{&message.UsageContent{Details: toUsageDetails(latestUsage)}},
+				Role:      message.RoleAssistant,
+				CreatedAt: time.Now(),
+			}, nil)
+		}
```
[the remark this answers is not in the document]
PratikDhanave: Good point — the trailing usage update now carries RawRepresentation set to the source chunk that produced the final usage, so no nil is appended to the message's RawRepresentation slice.

provider/geminiprovider/usage_streaming_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Move these tests to provider/geminiprovider/agent_test.go

qmuntal (changes requested):

provider/geminiprovider/usage_streaming_test.go
PratikDhanave: Done — moved into `agent_test.go`.

qmuntal (approved):

PR #510, opened 2026-07-16
Keep AdditionalProperties on every Responses output message

PratikDhanave:
## Summary

In the non-streaming Responses path (`provider/openaiprovider/responses.go`), when a response has more than one output message, `currentUpdate` is reset to a fresh `&agent.ResponseUpdate{}` for each message after the first. The per-message block then repopulates `MessageID`, `ResponseID`, `FinishReason`, `ContinuationToken`, `RawRepresentation`, `Role`, and `CreatedAt` — but **not** `AdditionalProperties`. So the second and later messages lose response-level properties such as `EndUserId` (from `resp.User`); only the first message (initialized at the top) and the aggregated response-level merge retain them.

## Fix

Repopulate `currentUpdate.AdditionalProperties = responsesPopulateAdditionalProperties(resp)` in the per-message block, so every output message carries the response-level properties.

## Public API

No exported symbols change.

## Tests

Adds `TestResponses_NonStreaming_AllMessagesKeepAdditionalProperties`: a response with two output messages and a response-level `"user"`. The second message's update reports an **empty `EndUserId` before this change** and the correct value after. (The existing `EndUserId` coverage only checked the aggregated `resp.AdditionalProperties`, which the response-level merge kept populated from the first message — masking the per-message drop.)

`go test ./provider/openaiprovider` passes; gofmt clean.


provider/openaiprovider/responses.go
```diff
@@ -835,6 +835,11 @@ func responsesProcessResponse(resp *responses.Response, seqNum int64, yield func
 			currentUpdate.RawRepresentation = out
 			currentUpdate.Role = message.Role(out.Role)
 			currentUpdate.CreatedAt = time.Unix(int64(resp.CreatedAt), 0)
+			// Repopulate response-level properties: currentUpdate is reset to a
+			// fresh value for each output message after the first, which would
+			// otherwise drop AdditionalProperties (e.g. EndUserId) on the second
+			// and later messages.
+			currentUpdate.AdditionalProperties = responsesPopulateAdditionalProperties(resp)
```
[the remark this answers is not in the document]
PratikDhanave: Done — moved the repopulation into the reset block so it only runs on a message transition; the first message keeps its initialized value.

provider/openaiprovider/responses_additionalprops_test.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Move these tests into responses_test.go

qmuntal (changes requested):

provider/openaiprovider/responses_additionalprops_test.go
PratikDhanave: Done — moved into `responses_test.go`.

PratikDhanave:
Both review points are addressed: the tests were moved into `responses_test.go`, and `AdditionalProperties` is now repopulated only in the per-message reset block (the first message keeps its initialized value, per the Copilot note). Rebased onto the latest `main` and green. Ready for another look whenever you have a moment — thanks!

PratikDhanave:
@qmuntal this one is ready for another look — both review points are addressed: the tests were moved into `responses_test.go`, and the `AdditionalProperties` repopulation was moved into the message-reset block so it only runs on a message transition (the first message keeps its initialized value). Functional CI is green. Thanks!

qmuntal (approved):

qmuntal:
Thanks! Needs rebase.

PR #511, opened 2026-07-16
Bounds-check fan-out Assigner indices into SinkIDs

PratikDhanave:
## Summary

`selectedTargetIDs` (workflow/internal/execution/edgerunner.go) indexes `edge.Connection.SinkIDs[id]` directly with each index yielded by a **caller-supplied** fan-out `Assigner` (attached via the public `WithEdgeAssigner`). The `Assigner` signature is `func(targetCount int, message any) iter.Seq[int]`, so an assigner that yields an out-of-range or negative index panics the workflow runtime with `index out of range` during message dispatch.

Framework-generated assigners (e.g. `SwitchBuilder`) stay in range, so this is a public-API hardening gap: a user assigner bug crashes the whole run rather than being contained.

## Fix

Skip any index outside `[0, len(SinkIDs))` instead of indexing blindly, so a misbehaving assigner cannot crash the runtime.

## Public API

No exported symbols change.

## Tests

Adds an internal test (`package execution`) with an assigner that yields a valid index, an out-of-range index (`5` for a 2-sink edge), a negative index, and another valid index. It **panics before this change** and returns the valid targets (`[a b]`) after.

`go test ./workflow/internal/execution ./workflow` passes; gofmt clean.


workflow/internal/execution/edgerunner.go
```diff
@@ -284,7 +284,11 @@ func selectedTargetIDs(edge workflow.Edge, envelope *MessageEnvelope) []string {
 	if edge.Assigner != nil {
 		targetIDs = make([]string, 0, len(edge.Connection.SinkIDs))
 		for id := range edge.Assigner(len(edge.Connection.SinkIDs), envelope.Message) {
-			targetIDs = append(targetIDs, edge.Connection.SinkIDs[id])
+			// Assigner is caller-supplied (WithEdgeAssigner); ignore any
+			// out-of-range index instead of panicking the workflow runtime.
+			if id >= 0 && id < len(edge.Connection.SinkIDs) {
```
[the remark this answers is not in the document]
PratikDhanave: Good call — added a nil-sequence guard (a nil iter.Seq panics when ranged) alongside the bounds check, plus a regression test `TestSelectedTargetIDs_AssignerReturningNilSeqIsTolerated`.

workflow/internal/execution/edgerunner_assigner_internal_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Rename this file to edgerunner_test.go

workflow/internal/execution/edgerunner_assigner_internal_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+package execution
```
qmuntal: I don't like testing internal behavior, can use use black box testing here?

qmuntal (changes requested):

workflow/internal/execution/edgerunner_assigner_internal_test.go
PratikDhanave: Done — renamed to `edgerunner_test.go` and rewrote as black-box tests (package `execution_test`) that drive the exported `EdgeRunner.PrepareDeliveryForEdge` with a fan-out edge, instead of calling the unexported `selectedTargetIDs` directly.

workflow/internal/execution/edgerunner_assigner_internal_test.go
PratikDhanave: Done — renamed to `edgerunner_test.go` and rewrote as black-box tests (package `execution_test`) that drive the exported `EdgeRunner.PrepareDeliveryForEdge` with a fan-out edge, instead of calling the unexported `selectedTargetIDs` directly.

qmuntal (approved):

PR #512, opened 2026-07-16
Skip nil-valued options in AllOptions instead of panicking

PratikDhanave:
## Summary

`AllOptions` panics with `"option type mismatch"` when a matched option's `Value()` cannot be asserted to `T` — whereas the sibling `GetOption` (a few lines up) returns `(zero, false)` gracefully for the identical case. A `WithTool(nil)` run option triggers this: its `Value()` is a nil `tool.Tool`, so the assertion `.(tool.Tool)` yields `ok == false` and `AllOptions` panics.

This is reachable through ordinary use — e.g. building options from a `[]tool.Tool` that contains a nil element. During `a.Run`, the `toolautocall` middleware calls `createToolsMap(agent.AllOptions(opts, agent.WithTool))` and every provider iterates `AllOptions(opts, WithTool)`, so the panic aborts tool collection everywhere.

## Fix

`continue` past an option whose value is absent for `T`, mirroring `GetOption`'s graceful handling and `agent.New`'s existing filtering of nil `Config.Tools`. A nil tool is simply skipped rather than crashing the run.

> Design note: I fixed this in `AllOptions` for consistency with `GetOption`. An alternative is to reject `nil` at `WithTool` construction (or guard in each collection loop) — happy to adjust to whichever you prefer.

## Public API

No exported signatures change. Behavior change is narrow: a type-mismatched/nil-valued option is skipped instead of panicking.

## Tests

Adds `TestAllOptions_NilToolIsSkippedNotPanic`: ranging `AllOptions` over `[]Option{WithTool(nil)}` **panics before this change** and yields nothing after.

`go test ./agent` passes; gofmt clean.


agent/options_test.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal, line 1: Rename this file to options_test.go.

qmuntal (changes requested):

agent/options_test.go
PratikDhanave: Done — renamed to `options_test.go`.

qmuntal (approved):

PR #522, opened 2026-07-17
Synchronize the in-memory checkpoint manager

PratikDhanave:
## Summary

`inMemoryManager` (workflow/checkpoint/manager.go) has an unsynchronized `Store map[string]*SessionCache` and per-session caches. The manager is **keyed by session ID**, and `NewInMemoryManager` documents no single-use restriction — so sharing one manager across concurrent workflow runs (distinct sessions committing at the same time) is a natural pattern. Under that use, concurrent `Commit` calls race on the map and can crash the process with `fatal: concurrent map writes`.

Under `-race`, a handful of concurrent commits reliably report data races on the `Store` access in `sessionStore`.

## Fix

Guard `Store` with a `sync.Mutex` in `Commit`, `Lookup`, and `RetrieveIndex`. The critical sections are pure in-memory map/cache operations (no I/O), so holding the lock across each method is cheap.

> **Design note:** I fixed this by making the in-memory manager safe for concurrent use, since its session-keyed design invites sharing. The alternative is to *document* it as not concurrency-safe, mirroring `FileSystemJSONStore`'s existing caveat, and leave callers to serialize. Happy to switch to the doc-only approach if you'd prefer the stores be consistent in that direction instead.

## Public API

No exported symbols change.

## Tests

Adds `TestInMemoryManager_ConcurrentSessions_NoRace` (internal test): 64 goroutines `Commit` and `RetrieveIndex` distinct sessions concurrently. It reports data races under `-race` **before** this change and passes after.

`go test -race ./workflow/checkpoint` passes; `go build ./...` and `go vet` clean; gofumpt clean.


workflow/checkpoint/manager_race_internal_test.go
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Move tests to manager_test.go.

qmuntal (changes requested):

workflow/checkpoint/manager_race_internal_test.go
PratikDhanave: Done — moved the concurrency test into manager_test.go. The in-memory manager's Commit/RetrieveIndex aren't part of the sealed Manager interface, so the test needs package-internal access; manager_test.go is now a white-box (package checkpoint) file. Still red-green under -race.

qmuntal (approved):

PR #530, opened 2026-07-17
Pass file-based script metadata to the script runner

PratikDhanave:
## Problem

`fsskills` file-based scripts hand the user-provided `ScriptRunner` a stripped `*Script`. In `newScript` (`source.go`):

```go
script := &skills.Script{Name: name}
return runner(ctx, owner, script, arguments)
```

But the *discovered* `Script` returned by `newScript` carries more:

```go
ParametersSchema: defaultFileScriptSchema,
AdditionalProperties: map[string]any{"fsskills.scriptFS": fsys},
```

So a `ScriptRunner` that inspects `script.ParametersSchema` or `script.AdditionalProperties["fsskills.scriptFS"]` (the script's backing `fs.FS`, e.g. to open and execute the file) receives empty values. The `ScriptRunner` signature deliberately passes a `*Script` rather than just a name, so runners are entitled to that metadata.

The in-tree subprocess runner only uses `script.Name` + the skill's root FS, so nothing broke — which is exactly why the contract violation went unnoticed.

## Fix

Populate the runner-facing `Script` with the same `ParametersSchema` and `AdditionalProperties` (sharing the single `additionalProperties` map) as the discovered `Script`.

## Test

`TestFileSource_Runner_ReceivesScriptMetadata` (black-box, in `source_script_test.go`) uses a runner that captures the received script and asserts `ParametersSchema` is non-empty and the `fsskills.scriptFS` entry is present. Fails on `main` (empty schema, missing key), passes with the fix.


PratikDhanave:
Rebased on latest main and resolved the conflict: the file-script Run path was refactored upstream into `newFileScriptRunFunc`, which still handed the runner a bare `&skills.Script{Name: name}`. I threaded the parameters schema and backing fs.FS through that helper so the fix (real metadata reaching the runner) is preserved on top of the refactor. Builds and tests pass; MERGEABLE again.

qmuntal (approved):

PR #541, opened 2026-07-17
Map text DataContent to an MCP text resource, not a binary blob

PratikDhanave:
## Problem

When exposing an agent as an MCP tool, `agentContentToMCPContent` converts a `message.DataContent` into an `mcp.EmbeddedResource`. Image and audio media types get dedicated content types, but every other media type — **including `text/*`** — fell through to the default case and was written to `Resource.Blob` as base64 binary:

```go
switch contentValue.TopLevelMediaType() {
case "image": ...
case "audio": ...
default:
    return &mcp.EmbeddedResource{Resource: &mcp.ResourceContents{
        URI: contentValue.Name, MIMEType: contentValue.MediaType, Blob: data, // text ends up here too
    }}
}
```

`mcp.ResourceContents` deliberately separates `Text` (readable) from `Blob` (base64 binary). The **reverse** mapping already relies on this — `mcpContentToAgentContent` branches on `Resource.Text != ""` to reconstruct a `TextContent`. So the writer never producing `Text` is an asymmetry: a text `DataContent` round-tripped through MCP becomes an unreadable binary blob, and MCP clients reading the resource treat text as binary.

## Fix

Add a `text` case that populates `Resource.Text` with the decoded string. Binary payloads (e.g. `application/octet-stream`) still use `Blob` — unchanged.

## Test

`TestAddToolReturnsTextDataAsTextResource` (black-box) returns a `text/plain` `DataContent` from a tool and asserts the resulting `EmbeddedResource` carries the text in `Resource.Text` (and not `Blob`). Fails on `main` (`Text = ""`, Blob populated), passes with the fix. The existing `TestAddToolReturnsBinaryDataAsEmbeddedResource` (octet-stream → Blob) is unchanged.


karianna:
@PratikDhanave merge conflict

qmuntal (approved):

PR #555, opened 2026-07-19
Clone shared runOptions before appending per-run options

PratikDhanave:
In `prepareRun`, the agent-configured run options were prepended with:

```go
if len(a.runOptions) != 0 {
    options = append(a.runOptions, options...)
}
```

`a.runOptions` is a field shared across **every** run of the `Agent`, and `New` builds it with `slices.Clone(cfg.RunOptions)` followed by a per-tool `append(..., WithTool(tool))`, which routinely leaves the slice with **spare capacity**. Because of that spare capacity, `append(a.runOptions, options...)` writes each run’s per-run options — the synthesized `WithSession(...)`/`noSessionProvided(true)`, or a caller’s `WithServiceID` — into `a.runOptions[len:cap]`, memory shared across all runs.

When a single `*Agent` (with ≥1 configured tool, the normal case) serves concurrent runs — the natural usage for an agent framework — this is a data race, and one run can observe another run’s clobbered session/service ID reaching the provider. Sequential runs mask it because length-bounded reads never see `runOptions[len:]`.

Every other analogous spot in the package already clones before appending a shared slice (`New` clones `RunOptions`, `Middlewares`, `prov.Middlewares`; `invoke` clones `options`); this prepend site missed it.

## Fix

```go
options = append(slices.Clone(a.runOptions), options...)
```

## Test

`TestAgent_ConcurrentRunsDoNotShareRunOptions` builds an agent with tools (so `runOptions` has spare capacity) and issues 64 concurrent runs, each with its own `WithServiceID`/`WithSession`. Under `-race` it reports a DATA RACE in `prepareRun` on the old code and passes with the fix.


qmuntal (approved):

qmuntal:
Will fix this issue in a PR I'm preparing.

PR #600, opened 2026-07-23
Replay assistant reasoning as Anthropic thinking blocks in buildMessageParam

PratikDhanave:
## What

`buildMessageParam` in `provider/anthropicprovider/agent.go` switches over the outbound `msg.Contents` but had no case for `*message.TextReasoningContent`. It handled `TextContent`, `FunctionCallContent`, `FunctionResultContent` and `DataContent` only.

The inbound path `buildBlock` produces exactly that type from Anthropic responses:
- `anthropic.ThinkingBlock` -> `TextReasoningContent{ProtectedData: Signature, Text: Thinking}`
- `anthropic.RedactedThinkingBlock` -> `TextReasoningContent{ProtectedData: Data}`

So when a prior assistant turn's history was sent back (e.g. multi-turn tool loops with extended thinking), the reasoning block and its cryptographic signature were silently dropped.

## Fix

Add a `*message.TextReasoningContent` case that replays the block, appended in iteration order (Anthropic emits reasoning before the rest of the turn):
- signature + text present -> `anthropic.NewThinkingBlock(ProtectedData, Text)`
- data only (text empty) -> `anthropic.NewRedactedThinkingBlock(ProtectedData)`
- both empty -> skipped, so streamed partials that never received a signature don't produce an invalid unsigned thinking block the API would reject.

## Why

Preserving thinking-block signatures across turns is required for Anthropic extended-thinking with tool use; the signature must round-trip or the API rejects the replayed assistant turn. This matches the .NET/Python SDKs, which reserialize reasoning content back into thinking / redacted_thinking blocks rather than dropping them, keeping cross-SDK conversation replay behavior aligned.

## Tests

Added black-box tests in the canonical `agent_test.go`, driven through the exported `Run().Collect()` path with the existing request-capturing httptest harness:
- `TestAssistantReasoningReplayedAsThinkingBlock` — asserts a thinking block carrying signature `sig123` is emitted first, ahead of the tool_use block.
- `TestAssistantRedactedReasoningReplayedAsRedactedThinkingBlock` — asserts a redacted_thinking block with its data is emitted.
- `TestAssistantUnsignedReasoningIsSkipped` — asserts an unsigned partial is dropped.

All three fail before the fix and pass after. `go build ./...`, `go vet` and `go test` on the package are green.


qmuntal (approved):

qmuntal:
Needs rebase.

PR #625, opened 2026-07-23
Stop serializing FunctionCall/Result Error to match .NET JsonIgnore Exception

PratikDhanave:
## What

Remove the `Error` field from `serializedFunctionCallContent` and `serializedFunctionResultContent`, and drop the marshal/unmarshal logic that persisted it as a string and reconstructed it via `errors.New`. The in-memory `Error error` field on `FunctionCallContent` and `FunctionResultContent` stays as informational-only state.

## Why

In .NET, both `FunctionCallContent.Exception` and `FunctionResultContent.Exception` carry `[JsonIgnore]` and are never serialized. Python behaves the same way. The Go port instead serialized the error to an `Error` string and, on unmarshal, fabricated a synthetic `errors.New(...)` that was never part of the original object graph. That diverges from the other SDKs and reconstructs an error value with a lost type/wrapping that the producer never intended to persist. This change aligns the wire format with .NET/Python: the exception is a local mapping/diagnostic detail, not serialized state.

## Tests

- Added `TestFunctionCallContent_ErrorNotSerialized` and `TestFunctionResultContent_ErrorNotSerialized` in `message/content_test.go`: marshaling a content with a non-nil `Error` produces JSON with no `Error` key, and unmarshaling JSON without an `Error` key yields a nil `Error` field.
- Updated `TestContentEncoding_Roundtrip` to drop the `Error` it previously set on `FunctionResultContent`, since that value is intentionally no longer round-tripped.
- `go build ./...`, `go vet ./message/...`, and `go test ./message/...` all pass.


gdams (approved):

gdams:
@PratikDhanave linter issue

gdams (approved):

gdams:
@copilot fix the failing linter checks

PR #628, opened 2026-07-23
Wire MCP prompts (ListPrompts/GetPrompt) through mcptool as promised by its package doc

PratikDhanave:
## What

The `tool/mcptool` package doc states it exposes an MCP server's tools **and prompts**, but only `ListTools` existed; a repo-wide grep for `ListPrompts`/`GetPrompt` returned zero matches outside that doc string. This wires the promised prompt capability:

- `ListPrompts(ctx, session)` enumerates the server's prompt descriptors via `session.ListPrompts`.
- `GetPrompt(ctx, session, name, args)` calls `session.GetPrompt` and converts each returned `mcp.PromptMessage` into a `*message.Message`, reusing the existing `mcpContentToAgentContent` mapping so text/image/audio/resource contents are preserved. The MCP role (`user`/`assistant`) is mapped onto the agent-framework role, and the raw `*mcp.PromptMessage` is retained on the message's `RawRepresentation`.

## Why

The underlying SDK (`mcp.ClientSession`) already supports `ListPrompts`/`GetPrompt`, so the capability was available but unwired, leaving the package doc's promise unmet. This mirrors the .NET and Python MCP integrations, where server prompts are first-class alongside tools: prompts are fetched and folded into the conversation as context messages rather than invoked as tools. Keeping the Go port aligned with that cross-SDK shape lets agents consume server-provided prompt templates the same way.

Unlike tools, MCP prompts are **not** `agent.Tool` instances — they materialize into conversation messages rather than being invoked — so `GetPrompt` returns `[]*message.Message` and `ListPrompts` returns the raw descriptors.

## How tested

Added black-box tests in the canonical `mcp_test.go` using the existing in-memory transport harness (`connectInMemory`):
- `TestListPromptsReturnsServerPromptDescriptors` registers a prompt via `AddPrompt` and asserts the descriptor (name, description, arguments) round-trips.
- `TestGetPromptMaterializesAgentMessages` asserts template arguments reach the server and that user/assistant prompt messages materialize into `[]*message.Message` with text content preserved through `mcpContentToAgentContent`.

`go build ./...`, `go vet ./tool/mcptool/...`, and `go test ./tool/mcptool/...` all pass.

## Open design questions

Opening as a draft because prompts don't fit the existing `agent.Tool` surface, so the public API shape is up for discussion:

- **API shape**: Is `GetPrompt` returning `[]*message.Message` and `ListPrompts` returning raw `[]*mcp.Prompt` the right altitude, or should there be a thin agent-facing prompt wrapper type (paralleling the tool wrapper) instead of leaking the SDK descriptor?
- **Scope**: This intentionally does not auto-register prompts anywhere or add pagination handling (single `ListPrompts` page). Should cursor-based pagination be handled here?
- **Role mapping**: MCP defines only `user`/`assistant`; unknown roles default to `user`. Confirm that matches the intended .NET/Python behavior.


qmuntal:
Won't add this new APIs for now. Thanks anyway!

PR #631, opened 2026-07-23
Add image-generation hosted tool to the OpenAI Responses provider

PratikDhanave:
## What

Adds `hostedtool.ImageGeneration`, a hosted-tool marker mirroring the existing `WebSearch`, `FileSearch`, `CodeInterpreter` and `MCPServer` types, and wires it into the OpenAI Responses provider so it is serialized into `params.Tools` as a `responses.ToolImageGenerationParam`.

## Why

The Responses provider already handles the **output** side of image generation: `responsesProcessResponse` and the streaming path decode `image_generation_call` items into `DataContent` (covered by `TestResponsesNonStreamingImageGenerationCall_MapsToDataContent`). The underlying SDK also supports the **request** side via `ToolUnionParam.OfImageGeneration`. But the tool switch in `responsesBuildCompletionParams` handled only `FuncTool`, `WebSearch`, `FileSearch`, `CodeInterpreter` and `MCPServer`, so there was no way to actually request image generation through the hostedtool abstraction — `params.Tools` never got an `image_generation` tool.

This closes that input-side gap and brings the Go port in line with the .NET/Python `HostedImageGenerationTool`, keeping the hosted-tool surface consistent across SDKs (cross-SDK alignment).

## How it works

- `ImageGeneration` carries its optional settings (`size`, `quality`, `background`, `output_format`, `partial_images`) via an `AdditionalProperties` map, exactly like `WebSearch`/`FileSearch`.
- The new `case *hostedtool.ImageGeneration` reads those keys, populates a `responses.ToolImageGenerationParam`, and appends a `responses.ToolUnionParam{OfImageGeneration: ...}`, following the existing `WebSearch`/`CodeInterpreter` cases.

## Tests

`TestResponsesImageGenerationTool_MapsToRequestParams` runs through the existing fake-transport harness: it sends `&hostedtool.ImageGeneration{AdditionalProperties: {size, quality}}` and asserts the outgoing request body contains a `image_generation` tool with `size`/`quality` populated (the server-side body assertion is the black-box proxy for `params.Tools`), then confirms the response still maps back to `DataContent`. It fails before the switch case is added (the request omits the tools array) and passes after. Pairs with the existing `TestResponsesNonStreamingImageGenerationCall_MapsToDataContent`.

`go build ./...`, `go vet` and `go test` pass for the changed packages.

## Open design questions

- **AdditionalProperties option surface**: the settings are passed as an untyped `map[string]any` (`size`/`quality`/`background`/`output_format`/`partial_images`) to match the `WebSearch`/`FileSearch` precedent. Is a loosely-typed map the desired long-term shape, or would typed fields on `ImageGeneration` (mirroring `FileSearch.MaximumResultCount`) be preferred?
- **Scope**: this PR maps `size`, `quality`, `background`, `output_format` and `partial_images`. The SDK also exposes `model`, `moderation`, `input_fidelity`, `action` and `input_image_mask` (inpainting). Happy to add those here or leave as follow-ups.
- **`partial_images` typing**: currently accepts `int`/`int64` from the map. If we keep the map surface, should JSON-number semantics (`float64`) also be accepted?


gdams:

qmuntal:
.NET doesn't provide an equivalent abstraction yet. Closing for now, thanks anyway!

PR #633, opened 2026-07-23
Add Workflow Mermaid/DOT visualization export (parity with .NET WorkflowVisualizer)

PratikDhanave:
## What

Adds `workflow/visualization.go` exporting two functions:

- `func ToMermaidString(wf *Workflow) string` — renders a built `Workflow` as a Mermaid `flowchart TD`.
- `func ToDotString(wf *Workflow) string` — renders the same graph as a Graphviz DOT `digraph`.

Both consume the existing reflection API (`ReflectExecutors`, `ReflectEdges`, `EdgeInfo`, `EdgeConnection`) — no production data structures were changed.

## Why (parity)

.NET's `WorkflowVisualizer.ToDotString` / `ToMermaidString` is a real, shipped capability, and `workflow/telemetry.go` already serializes the graph to JSON but stops short of emitting a human-renderable diagram. This closes that cross-SDK gap so Go workflows get the same visualization affordance qmuntal values for alignment. The renderers mirror the .NET semantics:

- Nodes are emitted per executor binding, with the start executor highlighted.
- Fan-out edges expand to one edge per target.
- Fan-in edges (`len(SourceIDs) > 1`) route through a synthesized junction node.
- Conditional edges (`HasCondition`) are dashed; `EdgeInfo.Label` is preserved.
- Nested sub-workflows (a binding whose `ExecutorBinding.RawValue` is a `*Workflow`, as produced by `inproc.BindSubworkflowAsExecutor`) render as Mermaid subgraphs / DOT clusters via recursion.

Node IDs are sanitized for Mermaid and labels are escaped for DOT. Output is deterministic (executors and edges sorted; fan-in edges deduplicated since `ReflectEdges` registers them under every source).

## Tests

`workflow/visualization_test.go` (black-box `workflow_test`, reusing the existing `newNoOpExecutor` harness):

- Builds start → fan-out → fan-in-barrier → conditional-labeled edge and asserts stable substrings in both outputs: start-node highlight, each executor node, dashed conditional marker, edge label, and the fan-in junction node.
- A nested sub-workflow case asserting the subgraph (Mermaid) / cluster (DOT) block.

`go build ./...`, `go vet ./workflow/`, and `go test ./workflow/...` all pass.

## Open design questions

Opening as a draft since this adds public API surface:

- **API shape / location** — free functions in `package workflow`, or methods on `*Workflow`, or a separate `workflow/visualization` subpackage? Kept them as package-level funcs to mirror the static `WorkflowVisualizer` helpers.
- **Renderer nuances** — fan-in is drawn through a synthesized `{{"fan-in"}}` junction (Mermaid) / `diamond` node (DOT); fan-out is left as parallel edges without a junction. Worth confirming this matches the .NET visual convention.
- **Styling** — highlight/dash choices are minimal and hardcoded; open to matching the exact .NET palette/shapes if there's a canonical set.
- **Follow-ups** — could also surface request ports and output executors as distinct node styles; deferred to keep this change surgical.


qmuntal:
Will not implement this for now. Closing to reduce noise, thanks!

PR #634, opened 2026-07-23
Disambiguate restored ToolApprovalRequest from FunctionCallContent by TypeID

PratikDhanave:
## What

`agentworkflow.requestDataContent` resolves the concrete request content carried by a `*workflow.ExternalRequest`. It probed `req.Data.As(*message.FunctionCallContent)` before `*message.ToolApprovalRequestContent`.

For a **live** request this is fine: the `PortableValue` holds a concrete pointer and the wrong-type probe is correctly rejected. For a request **restored from a checkpoint**, `req.Data` is a delayed-deserialized `PortableValue` (a `json.RawMessage`). A serialized `ToolApprovalRequestContent` unmarshals cleanly into an empty `FunctionCallContent` — they share field names, `encoding/json` ignores unknown fields, and `FunctionCallContent.UnmarshalJSON` reads then discards the `"type"` discriminator without a mismatch error. So a restored `ToolApprovalRequest` resolved to an **empty `FunctionCallContent`**.

The downstream effect: `normalizeResponseData` then failed to recognize the request as an approval, so the incoming `ToolApprovalResponseContent` was not re-keyed to the original request ID before being delivered back into the workflow (it kept the workflow-facing external request ID instead of the caller's original `RequestID`).

## Fix

Disambiguate on the request's persisted `req.Data.TypeID` before the `As` probes, so a restored request resolves to the same concrete type as a live one. The change is confined to `requestDataContent`; the existing probe order remains as a fallback for the zero/unknown TypeID case.

## Parity

`OutstandingRequests` are persisted and rehydrated by discriminated type in the .NET and Python SDKs; a checkpoint round-trip preserves whether a pending request is a function-call or a user-input/approval request. This aligns the Go port with that behavior — a resumed workflow sees the same request/response content types it saw before checkpointing.

## Tests

Added `TestNew_SerializedSessionResumesApprovalRequest` in the package's canonical `workflow_test.go`, mirroring the existing `TestNew_SerializedSessionResumesFromCheckpoint` (function-call) path but for approvals: it raises a `ToolApprovalRequest`, round-trips the session through JSON (forcing the delayed `PortableValue`), resumes with the matching `ToolApprovalResponseContent`, and asserts the workflow receives the original request ID `"req-1"`. It fails before the fix (workflow observes the external ID `approval-persisted_UserInput:req-1`) and passes after.

`go build ./...`, `go vet ./workflow/agentworkflow/...`, and `go test ./workflow/agentworkflow/...` all pass.


gdams (approved):

gdams:
@copilot rebase and fix the failing linter checks

PR #635, opened 2026-07-23
Surface Anthropic server_tool_use and web_search_tool_result response blocks

PratikDhanave:
## What

`buildBlock` in `provider/anthropicprovider/agent.go` only switched on `TextBlock`, `ThinkingBlock`, `RedactedThinkingBlock`, and `ToolUseBlock`. Anthropic's server-side tool response blocks (`server_tool_use` and `web_search_tool_result`, emitted when the model runs a hosted web search) hit no case and were silently dropped from the response.

This adds two cases to `buildBlock`:

- `anthropic.ServerToolUseBlock` -> a `*message.FunctionCallContent` (CallID/Name/Input mirroring the client-side `ToolUseBlock` case), so callers can observe the server tool request. It is appended in place rather than via the `functions` map, since that map is nil on the streaming `ContentBlockStartEvent` path and Anthropic has already executed the call.
- `anthropic.WebSearchToolResultBlock` -> a `*message.FunctionResultContent` keyed by `tool_use_id`, with each `web_search_result` mapped to a `*message.CitationAnnotation` (URL/Title), reusing the same citation shape as the existing `TextBlock` handling. The error content variant (`web_search_tool_result_error`) is guarded and surfaced via `Error` + the error code.

## Why

Cross-SDK parity: the Python client (`_chat_client.py`) already parses `server_tool_use` and `web_search_tool_result`. Without these cases the Go provider drops the entire server-side web-search interaction, so downstream consumers can neither see which query the model ran nor cite the sources it retrieved. This complements the request-side WebSearch mapping and is scoped to only the response-block region (`buildBlock`, not the tool loop) so it is reviewable independently.

## Testing

Added `TestServerToolUseAndWebSearchResultBlocks` to the canonical `agent_test.go`, black-box through `RunText().Collect()` with a fake transport that returns both blocks. It asserts a `FunctionCallContent` (correct CallID/Name/arguments) and a `FunctionResultContent` (correct CallID plus a populated `CitationAnnotation`) are produced rather than dropped. The test fails before the change and passes after. `go build ./...`, `go vet ./provider/anthropicprovider/...`, and `go test ./provider/anthropicprovider/...` all pass.

## Open design questions

- Scope: this PR handles only the response blocks. Request-side WebSearch tool configuration is a separate change.
- API shape: `server_tool_use` is surfaced as a `FunctionCallContent` and `web_search_tool_result` as a `FunctionResultContent` with citation annotations. Open to representing these as a dedicated content type if that better matches .NET/Python, but function call/result reuses existing shapes.
- Follow-ups: the streaming path accumulates these blocks (they flow through `buildBlock` on `ContentBlockStartEvent`); a dedicated streaming assertion could be added if desired.


qmuntal:
Too many parity issues, this needs some human love. Closing, thanks anyway!

PR #638, opened 2026-07-23
Add composable function-invocation middleware to the tool autocall loop

PratikDhanave:
## What

Adds a per-tool-call interception point to the automatic tool-calling loop.

- New `agent.FunctionInvocationContext` (`Function`, `Arguments`, `CallContent`, `Result`, `Terminate`, `Iteration`) and `agent.FunctionInvocationMiddleware` in `agent/middleware.go`.
- New `FunctionMiddleware []agent.FunctionInvocationMiddleware` on `toolautocall.Config`, threaded onto the autocall struct and composed around `tool.Call` inside `processFunctionCall`.

A middleware can:
- mutate `Arguments` before `next` runs (the tool sees the mutated args),
- replace the invocation result by returning a different value from `next`,
- set `Terminate` to stop the loop after the current round of results is returned.

The chain composes **outermost-first**, matching the run-level `runChain` composition: the first configured middleware runs first and wraps the rest. `Iteration` carries the zero-based tool-calling round.

## Why

The run-level `Middleware` only wraps a whole `RunFunc`; there was no way to intercept an individual tool invocation. This is a real gap versus the other SDKs:

- **.NET**: `FunctionInvocationDelegatingAgent` + `AIAgentBuilder.Use(function callback)` intercept each function call.
- **Python**: `FunctionMiddleware` does the same.

This brings the Go port to parity for that per-call hook while keeping the change surgical (no behavior change when `FunctionMiddleware` is empty).

## How it's tested

`TestFunctionInvoking_FunctionMiddleware` in the canonical `autocall_test.go` registers two middleware and, driving the existing `agenttest` harness, asserts:
- (a) arguments mutated before `next` are what the tool observes,
- (b) a result replaced after `next` is what the `FunctionResultContent` carries,
- (c) `Terminate` stops the loop (the provider is called exactly twice; the extra turn is never consumed),
- `Iteration` increments across rounds and the chain composes outermost-first.

`go build ./...`, `go vet ./agent/...`, and `go test ./agent/...` pass.

## Open design questions

Opening as a draft to align on the API before finalizing:

- **API shape / placement**: types live in the `agent` package (alongside run-level `Middleware`) and are wired only through `toolautocall.Config`. Should there also be an agent-builder-level `Use(...)` convenience mirroring .NET's `AIAgentBuilder.Use`, or is Config wiring sufficient?
- **Result source of truth**: currently the value *returned* from the chain becomes the function result (`Result` is populated for inspection). .NET mutates `context.Result` in place — should we standardize on one?
- **Concurrency under `AllowConcurrentInvocations`**: each call gets its own `FunctionInvocationContext`, so per-call state is isolated, but user middleware sharing state across parallel calls must synchronize themselves. Do we want to document this contract, guarantee ordering, or serialize middleware regardless of the concurrency setting (as .NET's delegating agent effectively does)?
- **Terminate semantics**: today `Terminate` stops the loop *after* the current round's results are streamed back. Should a terminating call instead short-circuit remaining calls in the same round?


qmuntal:
We should have this, but I don't like the approach taken. .NET supports hooking into tool calls via a delegating agent, which we map as a `agent.Middleware`. That middleware would replace the real tools by wrappers than have the desired lifecycle hooks. Closing for now.

PR #641, opened 2026-07-23
Add hostedtool.ToolSearch type and map it to OpenAI Responses ToolSearchToolParam

PratikDhanave:
## What

Add a `hostedtool.ToolSearch` type and wire it through the OpenAI Responses provider so a deferred tool-search request can be sent to the service.

- `tool/hostedtool/hostedtool.go`: new `ToolSearch` marker tool (fields `DeferredTools`, `Namespace`, `NamespaceDescription`, `AdditionalProperties`; `Name()` returns `tool_search`), with the usual `var _ tool.Tool` assertion.
- `provider/openaiprovider/responses.go`: new `case *hostedtool.ToolSearch` that builds a `responses.ToolSearchToolParam` and appends `responses.ToolUnionParam{OfToolSearch: &variant}`, mirroring the existing WebSearch/FileSearch/CodeInterpreter/MCPServer cases.

## Why

Parity with the .NET/Python stacks: `Microsoft.Extensions.AI.HostedToolSearchTool` exposes a hosted tool-search marker with deferred tools, but the Go port had no equivalent — the only `ToolSearch` occurrences were unwired auto-generated `internal/azaiprojects` constants. The OpenAI Go SDK (`responses.ToolSearchToolParam` / `ToolUnionParam.OfToolSearch`) already supports the wire shape, so nothing but the mapping was missing.

## How tested

`provider/openaiprovider/responses_test.go` adds `TestResponsesToolSearchTool_SerializesCorrectly`, a black-box test that drives a `*hostedtool.ToolSearch` through `RunText` against the existing fake-transport harness and asserts the serialized request contains a `tool_search` tool (with `description`/`execution`/`parameters`). It fails when the provider case is removed (the tool is silently dropped) and passes with it. `go build ./...`, `go vet`, and `go test ./provider/openaiprovider/... ./tool/hostedtool/...` are green.

## Open design questions (draft)

- **Field mapping scope.** The current OpenAI `ToolSearchToolParam` only exposes `Description`, `Execution`, and `Parameters`. `NamespaceDescription` maps to `Description`, and `execution`/`parameters` are read from `AdditionalProperties`. `DeferredTools` and `Namespace` are carried on the type for .NET parity but have no direct field in this SDK version — should they be folded into `Parameters`, or is carrying them for other providers/future SDK versions the right call?
- **API shape.** Should `Execution` be a first-class typed field on `hostedtool.ToolSearch` rather than an `AdditionalProperties` key, matching how other providers might consume it?
- **Follow-ups.** Wire the same tool through the response/streaming *output* path (tool_search_call items) if/when needed.


qmuntal:
.NET's HostedToolSearchTool  is still experiment. Would prefer not to port it yet. Thanks anyway!

PR #643, opened 2026-07-23
Add fileaccess harness context provider with shared-folder file tools rooted at a directory

PratikDhanave:
## What

Adds a new self-contained `agent/harness/fileaccess` package, wired like the sibling `agent/harness/todo` provider. `New(*Options)` returns a `Provider` backed by `agent.NewContextProvider`, and its `Provide` hook injects file tools plus instructions on each invocation.

The provider exposes the same six tools as the .NET `FileAccessProvider`:

- `file_access_read_file`
- `file_access_save_file`
- `file_access_list_files`
- `file_access_list_subdirectories`
- `file_access_search_files`
- `file_access_delete_file`

All operations go through a local-filesystem store rooted at a caller-granted `Options.RootDir` (not session state). `Options.ReadOnly` omits the save/delete tools, matching the .NET read-only shipping mode.

## Why

The Go harness tree (`agent/harness/`) had `agentmode`, `loop`, `todo`, `toolapproval`, and `toolautocall`, but no file-access counterpart, while the .NET SDK ships `FileAccessProvider` with exactly this tool set (`file_access_read_file` / `save_file` / `list_files` / `list_subdirectories` / `search_files` / `delete_file`) and a read-only mode. This closes that cross-SDK parity gap so Go agents can be granted a scoped shared folder.

## Safety

Every path is interpreted relative to the root. The store resolves paths with `filepath.Clean(filepath.Join(root, rel))` and rejects anything that is absolute or escapes the root via a prefix check, so `../outside.txt` and absolute paths are refused before touching the filesystem.

## Tests

`fileaccess_test.go` is black-box (`package fileaccess_test`) and reuses the same harness style as `todo_test.go` (`agenttest.CreateSession`, driving tools through the exported `Invoking` API). It covers: default tool set + instructions present; `ReadOnly` omitting save/delete; save->read round trip; `list_files` direct-children-only vs `list_subdirectories`; `search_files` regex across nested files; delete; and path-escape rejection (`../outside.txt`, absolute path, escaping write not creating a file).

`go build ./...`, `go vet ./agent/harness/fileaccess/...`, and `go test ./agent/harness/fileaccess/...` all pass.

## Open design questions

- **Scope**: The store is a minimal local-filesystem implementation held inside the package. Should it instead be an exported interface so callers can back it with other stores (blob, in-memory), mirroring any abstraction on the .NET side?
- **API shape**: Tool argument/return shapes (relative-path strings, `search_files` returning slash-separated relative paths matched against file *contents*) are chosen for parity; happy to align field names/semantics exactly with the .NET tool schemas if they differ.
- **Follow-ups**: The .NET provider ships read-only mode as an auto-approval rule via the tool-approval harness. This PR keeps the package self-contained (ReadOnly simply omits the mutating tools); wiring an explicit auto-approval rule through `toolapproval` could be a follow-up.


gdams:

qmuntal:
Too many parity findings. Porting this needs some human love. Closing for now.

PR #644, opened 2026-07-23
Add filememory harness context provider with session-scoped file_memory_* tools

PratikDhanave:
## What

Adds a new `agent/harness/filememory` package: a context provider that gives agents a small session-scoped virtual file store for persisting notes, scratch data, and intermediate results across turns of a long-running task.

The provider exposes five tools, constructed with `functool.MustNew` and returned from `Provide` as `agent.WithTool` options:

- `file_memory_write` — create or overwrite a file
- `file_memory_read` — read a file's content (empty string when absent)
- `file_memory_delete` — remove a file (reports whether it existed)
- `file_memory_ls` — list stored file paths
- `file_memory_grep` — find files whose content matches a regexp

State is held in a self-contained in-memory `AgentFileStore` and persisted through the session state bag (`loadStore`/`saveStore` keyed by a `stateKey`), so files survive across `Invoking` calls of the same session. A per-session mutex (weak-pointer identity keyed, with runtime cleanup) serializes concurrent tool invocations, matching the todo provider.

## Why

This ports the .NET `Harness/FileMemory/FileMemoryProvider` to Go. Go previously had no local file-memory harness — the only memory provider (`foundryprovider/memory.go`) is remote Foundry-store injection with no file tools and no local backend. The wiring mirrors the existing `agent/harness/todo` provider exactly (`agent.NewContextProvider` + `Provide` returning `WithTool` options + session-state persistence), keeping the harness family consistent and aligning the Go SDK with the .NET surface.

## How it's tested

Black-box tests in `filememory_test.go` (package `filememory_test`), reusing the same harness style as `todo_test.go` (`agenttest.CreateSession`, driving tools via `tool.FuncTool.Call`):

- `Invoking` returns the five `file_memory_*` tools as `WithTool` options plus non-empty instructions
- drive the store through write -> read -> ls -> grep and assert content and match filtering
- overwrite, delete (present and missing), empty-session listing, invalid-regexp error
- state persists across a second `Invoking` via the session state bag
- concurrent writes + reads under `-race`

`go build ./...`, `go vet ./agent/harness/filememory/...`, and `go test -race ./agent/harness/filememory/...` all pass.

## Open design questions

- **Scope**: keeps the store fully in-memory and self-contained (no dependency on a shared/not-yet-existing store package). Should `AgentFileStore` instead live in a shared package so multiple providers can reuse it?
- **API shape**: is a pluggable `AgentFileStore` (constructor option to inject a custom backend) desirable now, or should that wait for a concrete second backend (e.g. disk-backed)? Currently only the default in-memory store is wired.
- **Tool surface**: this ships the minimum useful set (write/read/delete/ls/grep). Follow-ups could add move/copy or content-append if the .NET provider exposes them.


gdams:

qmuntal:
Too many parity findings, this needs some human love. Closing, thanks anyway!

PR #652, opened 2026-07-23
Plumb EnableMessageInjection through provider AgentConfig so the message-injector loop is reachable

PratikDhanave:
## What

The `toolautocall` harness already fully implements `Config.EnableMessageInjection` and its `MessageInjector` loop (`inject.go`, wired in `autocall.go`, tested in `autocall_inject_test.go`, ported from .NET's `MessageInjectingChatClient`). However, no provider ever wired the flag: every provider built `toolautocall.Config` with only `Logger`/`LogSensitiveData`, so the message-injector loop was unreachable through the normal agent API.

This PR:
- Adds `EnableMessageInjection bool` to `agent.Config` (in `agent/agent.go`), beside the existing `DisableFuncAutoCall`.
- Passes `EnableMessageInjection: config.EnableMessageInjection` into each provider's guarded `toolautocall.New(toolautocall.Config{...})` literal: `openaiprovider` (chat + responses), `anthropicprovider`, `geminiprovider`, and `aguiprovider`.

Because each provider `AgentConfig` embeds `agent.Config` and forwards `config.Config` to `agent.New`, the new field flows to all providers uniformly. Previously the only way to enable injection was `DisableFuncAutoCall=true` plus hand-rebuilding the middleware, which loses provider defaults.

## Why (parity)

This mirrors the .NET `MessageInjectingChatClient` / `EnqueueMessages` surface, which is composable into the standard function-invocation pipeline. Exposing the flag on `AgentConfig` keeps the Go SDK aligned: tools can enqueue follow-up messages into the auto-call loop without opting out of the provider-managed middleware.

## Behavior change

None when the field is false (the default). The value is zero-valued for all existing callers, so the installed middleware behaves exactly as before.

## Tests

Added two black-box tests in the canonical `provider/openaiprovider/chat_test.go`:
- `TestChatMessageInjection_ToolInjectsMessage`: builds an agent with `AgentConfig.EnableMessageInjection=true` and a tool that calls `MessageInjectorFromContext(ctx).EnqueueMessages(...)`; asserts the injected message is forwarded to the provider on the following service round and the final answer is produced. This test fails before the wiring change and passes after.
- `TestChatMessageInjection_DisabledReturnsNilInjector`: negative case with the field false, asserting `MessageInjectorFromContext` returns nil and nothing is injected.

`go build ./...`, `go vet`, and `go test` (incl. `-race` on the new tests) pass for the changed packages.

## Open design questions

- **Scope**: This threads the existing harness flag through all five providers for consistency. Should the responses/agui paths be excluded if they are considered out of scope for the initial cut?
- **API shape**: `EnableMessageInjection` sits on `agent.Config` alongside `DisableFuncAutoCall`. Is that the desired home, or should message injection be surfaced as part of a broader composable-middleware config (cf. #638)?
- **Follow-ups**: Only the injector enable flag is plumbed; other `toolautocall.Config` knobs (e.g. `AllowConcurrentInvocations`, `MaximumIterationsPerRequest`) remain provider-internal. Happy to plumb those in a follow-up if that is the direction.


qmuntal:
I'm still not sure how to expose autocall configuration to the agent constructors. Closing for now, thanks anyway!

PR #653, opened 2026-07-23
Add WhenUpdatesCompleted to Foundry MemoryProvider to await async memory extraction

PratikDhanave:
## What

Adds `WhenUpdatesCompleted(ctx, pollingInterval)` to the Foundry `MemoryProvider` so callers can await asynchronous memory extraction before searching.

`store()` previously discarded the poller returned by `BeginUpdateMemories`, making memory updates fire-and-forget with no way to know when they land. This retains the most recent poller on the provider (guarded by a `sync.Mutex`) and drives it to completion on demand via `runtime.Poller.PollUntilDone`. `store()` stays non-blocking, so only opt-in callers pay the polling cost.

- Returns `nil` when no update is pending or the update completed successfully.
- Returns a non-nil error when the update failed.
- Clears the pending update on success so repeated calls are no-ops.

## Why (parity)

Memory extraction on Foundry is a long-running operation. The .NET/Python agent-framework Foundry memory providers submit the update and expose a way to await the extraction result (the LRO poller / update-result endpoint) so a subsequent `SearchMemories` sees the freshly written memories. The Go port kept the fire-and-forget submit but never surfaced the completion handle, so tests and callers had no deterministic way to sequence store -> search. This closes that gap while keeping the hot path unchanged.

Not a duplicate of #627/#619/#617.

## How tested

`go build ./...`, `go vet ./provider/foundryprovider/...`, `go test ./provider/foundryprovider/...` all pass. New tests in `memory_test.go` use the existing `recordingTransport` to:
- script `BeginUpdateMemories` (202 + `Operation-Location`) then a status sequence `queued -> completed`, asserting `WhenUpdatesCompleted` returns nil and clears the pending update (a second call issues no further requests);
- script a `failed` status, asserting a non-nil error;
- assert the no-pending case returns nil immediately without any request.

## Open design questions

- **API shape**: method exposes `pollingInterval` and relies on the caller's `ctx` for the overall deadline. Should it instead take a full `runtime.PollUntilDoneOptions`, or default the interval?
- **Scope**: only the single most-recent update is tracked. Concurrent/overlapping updates keep only the latest poller. Is per-scope tracking wanted, or is latest-wins sufficient?
- **Terminal states**: `PollUntilDone` treats `completed` as success and `failed` as error. Foundry also has a `superseded` terminal state (an update replaced by a newer one); the generic poller does not treat it as terminal. If awaiting a superseded update matters, a follow-up could poll `GetUpdateResult` directly and treat `superseded` as success. Flagging as a follow-up rather than expanding scope here.


qmuntal:
@PratikDhanave fix the identified parity gaps please.

qmuntal:
Foundry memory is still a bit experiment, I prefer not to expose more APIs related to it yet. Thanks anyway!

PR #659, opened 2026-07-23
Wire MCP client options in mcptool.Connect: server sampling handler and list_changed notifications

PratikDhanave:
## What

`mcptool.Connect` constructed the MCP client with a nil `*mcp.ClientOptions`, so no client-side handlers were ever registered. Two go-sdk capabilities were therefore dead:

- **Sampling.** `ClientOptions.CreateMessageHandler` auto-advertises the sampling capability when non-nil and dispatches `sampling/createMessage`. With nil options the client advertised no sampling and rejected every server-initiated `createMessage`, so an MCP server that delegates sub-reasoning to the host model could not work.
- **List changed.** `ClientOptions.ToolListChangedHandler` / `PromptListChangedHandler` are dispatched for `notifications/tools/list_changed` and `notifications/prompts/list_changed`, but with nil options those notifications were dropped, so a server adding/removing tools mid-session could not refresh the caller's static tool set.

Both gaps live in the same `Connect` construction site. This change builds a non-nil `ClientOptions` from additive functional options and keeps `Connect` behaving exactly as before when called with no options:

- `WithSampling(client, approve, maxTokens)` sets `CreateMessageHandler`. It **denies by default** when no approver is supplied, clamps the server-requested `MaxTokens`, translates the request messages and `SystemPrompt` through the existing `mcpContentToAgentContent` mapper, invokes the caller's chat client, and returns the reply as a `CreateMessageResult{Role: assistant, StopReason: endTurn}` built with the existing `agentContentToMCPContent` mapper.
- `WithToolListChanged(cb)` / `WithPromptListChanged(cb)` set the notification handlers to invoke the caller callback (typically re-running `mcptool.ListTools`).

## Why (cross-SDK parity)

Python wires `ClientSession(sampling_callback=...)` and denies sampling by default unless the host explicitly opts in; .NET wires a sampling `IChatClient` on the MCP client. The Go port previously exposed neither, so host-in-the-loop sampling and dynamic tool/prompt refresh were unreachable. The deny-by-default policy here mirrors the Python client so a server cannot silently borrow the host model.

## Tests

Added to the canonical `tool/mcptool/mcp_test.go`, black-box via the in-memory transport already used by the suite:

- Sampling deny-by-default: the server issues `CreateMessage`; with `WithSampling` and no approver the call returns an error and the stub chat client is never invoked.
- Sampling approve-all: with an approving callback the stub is invoked, `MaxTokens` is clamped, the system prompt and user message are translated in order, and the reply is returned as an assistant/endTurn result.
- Tool list changed: `WithToolListChanged` fires the callback after `server.AddTool`, and a subsequent `ListTools` returns the newly added tool.

`go build ./...`, `go vet ./tool/mcptool/...`, and `go test -race ./tool/mcptool/...` pass.

## Open design questions

- **API shape.** Sampling is exposed through a minimal `SamplingChatClient` interface (`GetResponse(ctx, []*message.Message, maxTokens) (*message.Message, error)`) rather than binding to a concrete provider, to keep `mcptool` decoupled. Would a tie-in to the agent/provider chat-client type be preferred instead?
- **Scope.** This wires sampling + tool/prompt list-changed only. Elicitation, resource list-changed/updated, and logging handlers are further `ClientOptions` fields left as follow-ups. Prompts (#628), approval mode (#616), and name normalization (#611) are handled separately.
- **Reply reduction.** `CreateMessageResult.Content` carries a single content block; the handler uses the first non-nil content of the reply. Should multi-content replies be coalesced differently?


gdams:
@PratikDhanave needs a rebase

qmuntal:
We try to follow .NET MAF, and it still doesn't implement these new APIs. Closing for now.

PR #666, opened 2026-07-23
Emit gen_ai.client.token.usage and operation.duration metrics from the OTel middleware

PratikDhanave:
## What

The OTel middleware previously recorded token counts purely as span attributes (`gen_ai.usage.*`) via `setUsage`, and never measured run duration at all. It imported only `otel/{attribute,codes,trace}` and held a lone `tracer` — no Meter, no histograms.

Span attributes describe a single run and are not aggregatable across runs, so a spend/latency dashboard summing across many invocations saw nothing from the Go port. This change adds a Meter alongside the tracer and emits the two histograms the OTel GenAI semantic conventions define:

- `gen_ai.client.token.usage` (unit `{token}`) — two data points per run, one tagged `gen_ai.token.type=input` (`InputTokenCount`) and one `output` (`OutputTokenCount`).
- `gen_ai.client.operation.duration` (unit `s`) — one sample per run, measured from before `span.Start`.

Every data point carries the run's identifying attributes (`gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.agent.name`), plus `error.type` when the run faulted. Recording is factored into a `recordMetrics` helper (mirroring `setUsage`) shared by both the normal post-loop path and the early-return path. Histogram construction errors are tolerated: `NewMiddleware` still returns a working tracer-only middleware and `recordMetrics` no-ops on a nil histogram.

## Why (cross-SDK parity)

This aligns the Go port with the OTel GenAI semantic conventions and the Python/.NET references, which both emit these two histograms. Cross-language dashboards can now aggregate Go runs alongside the rest. Span-attribute accounting stays exactly as-is; metrics complement it (single-run detail vs. cross-run aggregation) rather than replacing it. Not a duplicate of the span PRs (#655/#646) or usage-accounting PRs (#594/#609/#551-554).

## Tests

`provider/otelprovider/otel_test.go` gains two black-box tests driving an in-memory `sdkmetric.ManualReader`:

- `TestOtel_Run_RecordsUsageAndDurationMetrics` — runs the middleware over a fake `RunFunc` yielding `UsageContent` (100 input / 50 output), collects `ResourceMetrics`, and asserts the two `gen_ai.client.token.usage` data points (input=100, output=50 by `gen_ai.token.type`) with the shared attributes, plus one `gen_ai.client.operation.duration` sample.
- `TestOtel_Run_MetricsIncludeErrorType` — asserts `error.type` is present on the duration metric when the run faults.

Both fail before the change and pass after. `go build ./...`, `go vet ./provider/otelprovider/...`, and `go test ./provider/otelprovider/...` are green.

## Open design questions

- **Scope:** should these live in the existing tracing middleware (as here) or a separate metrics middleware/config toggle? Kept together since they share the accumulated usage and the run boundary.
- **API shape:** `MiddlewareConfig` is unchanged — the Meter reuses `SourceName` for its instrumentation scope, matching the tracer. Open to a separate meter name if preferred.
- **Follow-ups:** `AdditionalCounts` and cache/reasoning sub-counters are not yet emitted as metric data points (the conventions only define input/output token-type dimensions); could be added if there's appetite.


gdams:
@PratikDhanave conflicts

qmuntal:
Thanks for identifying this gap. I've taken another approach in this PR: https://github.com/microsoft/agent-framework-go/pull/922.

PR #739, opened 2026-07-24
Do not open a spurious empty WorkflowRun span on no-work wakeups in the streaming run loop

PratikDhanave:
## What

In `workflow/internal/execution/eventstream.go`, `streamingRunEventStream.runLoop()` opened a WorkflowRun span on every wakeup: it called `telemetry.StartWorkflowRun` + `runActivity.AddEvent(EventWorkflowStarted)` unconditionally, and `AddEvent(EventWorkflowCompleted)` + `End()` unconditionally. Only the actual superstep work and the `StartedEvent` were guarded by `if s.stepRunner.HasUnprocessedMessages()`, with the existing comment noting that events should be emitted "only when there's actual work to process, to avoid spurious events on no-work loop iterations." The telemetry span was left outside that guard.

No-work wakeups are reachable in normal operation:
- the loop is signaled when only `HasUnservicedRequests()` is true (no messages queued), and
- checkpoint restore always signals the run loop even when no messages are queued.

Each such wakeup therefore produced a zero-superstep `workflow_invoke` span with no `executor.process` children.

This change moves `StartWorkflowRun` + `EventWorkflowStarted` inside the `HasUnprocessedMessages()` branch, co-located with the `StartedEvent` emission, and finalizes the span (`EventWorkflowCompleted` + `End()`) only when a run activity was actually started, guarding on a nil `runActivity`. No-work iterations now create no span.

## Why

This mirrors the sibling **lockstep** implementation (`lockstepRunEventStream`) in the same file, which already opens the run span lazily via `startRunActivity()` and guards finalization with `if runActivity != nil`. Aligning the streaming path removes an inconsistency between the two execution modes and keeps WorkflowRun span semantics matching the .NET/Python behavior, where a run/invoke activity corresponds to an actual unit of work rather than an idle wakeup. It also brings the telemetry in line with the `StartedEvent` guard that already lived in this loop.

## Tests

Added `TestStreamingRunEventStream_NoWorkWakeupDoesNotOpenWorkflowRunSpan` to the package's canonical `eventstream_test.go`. It builds a telemetry-enabled workflow with an in-memory recording tracer, starts the streaming run loop, and triggers a no-work wakeup (`SignalInput` while `HasUnprocessedMessages()` reports false, as in a checkpoint restore whose only state is an unserviced request). It synchronizes on the loop's internal halt signal, then asserts zero `workflow_invoke` spans. The test fails before the fix (count = 1) and passes after.

`go build ./...`, `go vet ./workflow/internal/execution/...`, and `go test ./workflow/internal/execution/... -race` all pass.


workflow/internal/execution/eventstream_test.go
```diff
@@ -55,6 +57,46 @@ func TestStreamingRunEventStream_ErrorEventCancelsRunLoop(t *testing.T) {
...
+	stream := newStreamingRunEventStream(runner, false)
+
+	stream.Start()
+	defer stream.Stop()
+	waitForEventHandlerCount(t, runner.outgoingEvents, 1)
+
+	// Trigger a no-work wakeup: the run loop is signaled but
+	// HasUnprocessedMessages reports false, mirroring a checkpoint restore
+	// whose only state is an unserviced request. Such an iteration must not
+	// open a WorkflowRun span, matching the lockstep implementation.
```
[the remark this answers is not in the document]
PratikDhanave: Good catch — fixed in 9b95de4: the runner returns false for HasUnservicedRequests(), so I reworded the comment to describe what the test actually does (signal the loop while HasUnprocessedMessages() reports false) rather than the checkpoint-restore-with-unserviced-request scenario.

workflow/internal/execution/eventstream.go
```diff
@@ -176,8 +176,8 @@ func (s *streamingRunEventStream) runLoop() {
 		default:
 		}
 
-		cycleCtx, runActivity := telemetry.StartWorkflowRun(ctx, workflowMetadata(wf, s.stepRunner.SessionID()))
-		runActivity.AddEvent(observability.EventWorkflowStarted)
+		cycleCtx := ctx
+		var runActivity *observability.Activity
```
qmuntal: These declarations are not necessary if the runActivity.AddEvent and runActivity.End calls are moved inside the s.stepRunner.HasUnprocessedMessages block.

qmuntal (dismissed):

workflow/internal/execution/eventstream.go
PratikDhanave: Done in 992ab19 — moved the completion (AddEvent(EventWorkflowCompleted) + End()) inside the HasUnprocessedMessages block, so runActivity is now declared inside the branch and the outer declaration plus the `runActivity != nil` guard are gone.

gdams (approved):

PR #750, opened 2026-07-24
Add an example demonstrating imperative in-executor human-in-the-loop via Context.PostRequest

PratikDhanave:
## What

Adds `examples/03-workflows/human-in-the-loop/imperative_request`, a runnable sample where a custom `*workflow.Executor` raises a human-in-the-loop request imperatively from inside a handler via `Context.PostRequest` and consumes the matching `*workflow.ExternalResponse` itself. It is registered in `cmd/verifyexamples/examples.go`.

## Why

`Context.PostRequest` is documented in `workflow/workflow.go` as a first-class user primitive and is exercised by `TestPostRequestFromExecutor`, but no committed example showed it — the only HITL sample (`human_in_the_loop_basic`) uses the higher-level `workflow.RequestPort` binding. This sample fills that gap and is framed explicitly as the low-level counterpart to the request-port binding, mirroring the .NET/Python guidance where the request-port abstraction and the imperative post-request path are presented as two ways to reach the same external-request/response protocol. Surfacing both keeps the Go examples aligned with the other SDKs.

## How it works

The executor registers two raw handlers via `RouteBuilder`: on the initial `string` input it builds `workflow.NewExternalRequest` and calls `wctx.PostRequest`; on `*workflow.ExternalResponse` it reads `resp.Data.As(port.Response)` and yields the greeting. The driver runs with `inproc.Default.Run`, drains events to catch `workflow.RequestInfoEvent`, answers with a fixed value via `run.Resume`, and drains the resulting `OutputEvent`. The shape follows `TestPostRequestFromExecutor`.

## Testing

- `go build ./...`
- `go vet` on the new example and `cmd/verifyexamples`
- Ran the example: it prints the raised prompt and completes with `Hello, Ada!`
- Deterministic (fixed answer), registered in verifyexamples with a `MustContain` assertion


imperative_request
gdams, line 1: Looks like you checked this binary in by accident

gdams (approved):

PR #863, opened 2026-08-18
[dotnet-port-fixes] Execute invocable tool siblings

michelle-clayton-work:
## Summary

Updated `agent/harness/toolautocall` so invocable local tool calls still run when the same assistant turn also includes declaration-only schema tools. The middleware now executes the invocable siblings, leaves declaration-only calls actionable for the caller, and marks only the executed local calls informational-only. Added a regression test covering mixed-call ordering.

## Ported .NET PRs

- microsoft/agent-framework#7388 - Store executable function calls bypassed by declaration-only tool calls. Ported the behavior correction from upstream commit [`aaaa56bc60b6703b305664f196e74b284e425918`](https://github.com/microsoft/agent-framework/commit/aaaa56bc60b6703b305664f196e74b284e425918), based on `dotnet/src/Microsoft.Agents.AI/ChatClient/InvocableFunctionBypassingChatClient.cs` and `dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/InvocableFunctionBypassingChatClientTests.cs`.

## Breaking Changes

No.

## Tests and Examples

- `go test ./agent/harness/toolautocall`
- Added `TestFunctionInvoking_MixedInvocableAndNonInvocableSchemaToolExecutesInvocableSibling`
- No examples changed

## Notes

- Kept the fix internal to `agent/harness/toolautocall`; no exported Go API changed.
- Preserved existing `TerminateOnUnknownCalls` behavior for unknown tools and only realigned mixed invocable/declaration-only schema-tool handling.

> Generated by [.NET to Go Fixes and Test Porting Agent](https://github.com/microsoft/agent-framework-go/actions/runs/31766010485) · gpt54 · 212.2 AIC · ⌖ 12.4 AIC · ⊞ 24.2K · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fagent-framework-go+%22gh-aw-workflow-id%3A+dotnet-port-fixes-nightly%22&type=pullrequests)

Closes #842

PR #872, opened 2026-08-19
Fix workflow port prompts

michelle-clayton-work:
Fixes the wording to better clarify what should be classified as api and fixes 

in response to: **https://github.com/microsoft/agent-framework-go/pull/863#pullrequestreview-4971285592**


qmuntal (approved):
Thanks!

PR #878, opened 2026-08-20
Clarify agent input ownership

qmuntal:
## Summary

- Establish a read-only ownership contract for provider run functions, middleware, context/history providers, and provider session creation.
- Move cloning to actual mutation boundaries so caller- and callback-owned message/option slices are preserved while pass-through paths avoid unnecessary copies.
- Make message cloning own top-level slice/map containers and preserve approval replay after serialization by correlating paired responses by request ID.

The primary intent of this PR is to make ownership explicit and consistent: incoming message and option slices, plus existing messages, are borrowed read-only; code that modifies derived input or source attribution must own its copy. Allocation cleanup follows that contract rather than weakening it.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -16,6 +16,8 @@ import (
 )
 
 // RunFunc is the provider function that executes an agent invocation.
+// Implementations must treat the message and option slices, and existing
+// messages, as read-only. Clone them before making changes.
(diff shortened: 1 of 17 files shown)


qmuntal:
@copilot fix codereview findings

gdams (approved):

PR #895, opened 2026-08-24
Prevent duplicate Agent Framework user agents

qmuntal:
## Summary

- make Agent Framework User-Agent stamping idempotent
- normalize bare, hosted, stale, and duplicate `agent-framework-go` products to one current product
- add regression coverage for map headers, HTTP headers, and hosted upgrades

## Testing

- `go test -count=1 ./internal/telemetry ./provider/openaiprovider ./provider/foundryprovider`

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -57,12 +57,7 @@ func PrependAgentFrameworkToUserAgent(headers map[string]string) map[string]stri
 	if headers == nil {
 		headers = map[string]string{}
 	}
-	userAgent := userAgent()
-	if existing := headers[userAgentKey]; existing != "" {
-		headers[userAgentKey] = userAgent + " " + existing
-	} else {
-		headers[userAgentKey] = userAgent
-	}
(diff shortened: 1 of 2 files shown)


internal/telemetry/telemetry.go
```diff
@@ -74,15 +69,50 @@ func PrependAgentFrameworkToHTTPHeader(headers http.Header) http.Header {
...
-	} else {
-		headers.Set(userAgentKey, userAgent)
-	}
+	headers.Set(userAgentKey, prependUserAgent(headers.Get(userAgentKey)))
 	return headers
 }
 
+func prependUserAgent(existing string) string {
+	frameworkUserAgent := userAgent()
+	products := strings.Fields(existing)
```
michelle-clayton-work, line 78: Based on how strings.Fields works -- would this case `"my-app/1.0 (compatible; agent-framework-go/v1.0)"` pass?

qmuntal: Yes, we could fix that, but it would complicate code. And the current implementation is taken from .NET, which also has this issue 😸

gdams (approved):

Issue #925, opened 2026-08-27
Add scheduled end-to-end validation for Go examples


michelle-clayton-work:
## Problem

`cmd/verifyexamples` runs the Go examples and checks their output, but no CI workflow currently runs it against live model providers.

Unit tests use mocked providers, so they do not catch authentication failures, provider API changes, or other problems that only appear in a real model call.

PR #431 added `verifyexamples` and left scheduled CI as follow-up work once provider access was available.

## Proposed work

Add a scheduled and manually dispatched GitHub Actions workflow that:

- Runs a small, stable set of examples through `cmd/verifyexamples`.
- Writes the Markdown report to the GitHub job summary.
- Uploads the detailed report as an artifact.
- Uses short-lived authentication where possible.
- Keeps API keys and tokens out of the repository.
- Starts as an informational check rather than a required PR check.

## Provider investigation

Start by checking whether the GitHub Copilot provider can use authentication already available to GitHub Actions. Copilot-only coverage would still test real agent calls without requiring accounts for every supported provider.

`verifyexamples` currently uses Foundry for semantic output verification. If Copilot is the first provider tested, decide whether to:

- Use deterministic output checks.
- Add Copilot-backed semantic verification.
- Keep Foundry as the verifier and configure Azure OIDC.

For broader provider coverage, ask the Agent Framework team how they provision and manage credentials for their .NET and Python integration workflows. Anthropic and Gemini may require separate paid credentials.

## Done when

- A scheduled and manually dispatched workflow runs selected examples.
- At least one live provider path runs end to end.
- Missing credentials fail or skip with a clear reason.
- The workflow publishes a report even when an example fails.
- Authentication and ownership are documented.
- No reusable credential is committed to the repository.

## References

- https://github.com/microsoft/agent-framework-go/pull/431
- https://github.com/microsoft/agent-framework/blob/main/.github/workflows/dotnet-verify-samples.yml
- https://github.com/microsoft/agent-framework/blob/main/.github/workflows/python-sample-validation.yml

PR #931, opened 2026-08-27
ci: verify examples and Copilot integration tests

michelle-clayton-work:
## Summary

Add scheduled and manually dispatched CI that runs every example supported by the configured credentials. Add GitHub Copilot provider integration tests to the normal test workflow.

The first provider covered is GitHub Copilot. Both workflows use the short-lived `${{ github.token }}` with `copilot-requests: write`, so they require no repository secret, API key, Foundry project, or Azure setup.

## Example verification

`live-examples.yml` runs the existing `cmd/verifyexamples` command without a name filter. Examples for providers without configured credentials skip with their existing reasons.

Two fixed-input number-guessing examples now use their existing deterministic output checks instead of requiring Foundry semantic grading.

A validated run executed 24 available examples and skipped 83 unavailable examples with explicit reasons:
https://github.com/microsoft/agent-framework-go/actions/runs/33197392893

## Copilot integration tests

Port all tests from the .NET Agent Framework `GitHubCopilotAgentTests.cs` into `provider/copilotprovider/agent_e2e_test.go`:

- nonstreaming response
- streaming response
- function-tool invocation
- approval-required function tool
- same-session context
- cross-client session resume
- shell command permission
- URL fetch permission
- local MCP filesystem server
- remote MCP server, explicitly skipped to match its upstream disabled category

These tests run in a dedicated Linux job in `.github/workflows/test.yml`. Ordinary test runs skip them. GitHub Actions runs them when `GITHUB_TOKEN` is available, and developers can opt in locally with `RUN_COPILOT_INTEGRATION_TESTS=true`.

The integration job does not run for fork pull requests because it executes checked-out code with Copilot access.

## Authentication and safety

- Checkouts do not persist Git credentials.
- Copilot receives the job-scoped GitHub Actions token only in live test steps.
- Integration tests disable config discovery, file hooks, git operations, session storage, and skills.
- The normal test workflow's conclusion job includes the Copilot integration job.
- The examples workflow uploads reports even when verification fails.

## Validation

- `go test -count=1 ./...`
- `go vet ./cmd/verifyexamples ./provider/copilotprovider`
- YAML parsing and `git diff --check`
- Standard Tests workflow, including Copilot integration: https://github.com/microsoft/agent-framework-go/actions/runs/33421954501

Nine active Copilot integration tests passed. The remote MCP case skipped as intended.

## Upstream reference

- https://github.com/microsoft/agent-framework/blob/main/dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.IntegrationTests/GitHubCopilotAgentTests.cs

- https://github.com/microsoft/agent-framework-go/issues/925


cmd/verifyexamples/examples.go
```diff
@@ -717,6 +719,13 @@ var agentsExamples = []ExampleDefinition{
 			"The output should not contain error messages or stack traces.",
 		},
 	},
+	{
+		Name:            "02_agents_providers_github_copilot_function_tool",
+		ProjectPath:     "examples/02-agents/providers/github-copilot/function_tool",
+		IsDeterministic: true,
+		MustContain:     []string{canaryResult},
+		MustNotContain:  []string{"panic:", "Permission denied"},
+	},
```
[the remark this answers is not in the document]
michelle-clayton-work: This example does not require COPILOT_GITHUB_TOKEN. The workflow supplies it as an explicit CI credential, but when it is absent the Copilot SDK defaults to the locally signed-in Copilot or gh user. This matches the existing GitHub Copilot example, which also has no RequiredEnvironmentVariables entry.

Adding it as required would cause verifyexamples to skip valid local runs for signed-in developers. The scheduled workflow always supplies the short-lived GitHub Actions token, as confirmed by the passing live run, so I’m leaving this unchanged.

provider/copilotprovider/agent_e2e_test.go
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal, line 1: Port all tests from https://github.com/microsoft/agent-framework/blob/main/dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.IntegrationTests/GitHubCopilotAgentTests.cs.

.github/workflows/live-examples.yml
```diff
@@ -0,0 +1,99 @@
...
+          set -e
+
+          if [[ -f live-examples.md ]]; then
+            cat live-examples.md >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "The example verifier exited before it could write a report." >> "$GITHUB_STEP_SUMMARY"
+          fi
+          exit "$status"
+
+      - name: Run Copilot integration tests
```
qmuntal: The integration tests should run as part of the test workflow, they are orthogonal to the examples.

qmuntal (changes requested):
Like the direction 😸

provider/copilotprovider/agent_e2e_test.go
michelle-clayton-work: Done. I ported all tests from GitHubCopilotAgentTests.cs. Nine run in CI, and the remote MCP test is explicitly skipped to match its upstream IntegrationDisabled category.

The full Copilot integration job passed:
https://github.com/microsoft/agent-framework-go/actions/runs/33421954501

.github/workflows/live-examples.yml
michelle-clayton-work: Done. The Copilot integration tests now run in a dedicated job in test.yml. live-examples.yml only runs verifyexamples.

The standard Tests workflow, including the Copilot integration job and final conclusion, passed:
https://github.com/microsoft/agent-framework-go/actions/runs/33421954501

cmd/verifyexamples/examples.go
```diff
@@ -708,14 +708,13 @@ var agentsExamples = []ExampleDefinition{
 		SkipReason:                   "Spawns a local stdio MCP server subprocess (go run) that requires live Foundry credentials.",
```
qmuntal, line 708: Why the changes in this file?

.github/workflows/live-examples.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Check results
+        if: always()
+        shell: bash
+        env:
+          EXAMPLES_OUTCOME: ${{ steps.examples.outcome }}
+        run: |
+          if [[ "$EXAMPLES_OUTCOME" != "success" ]]; then
+            echo "Example verification failed."
+            exit 1
+          fi
```
qmuntal: Add a new line at the end of the file.

provider/copilotprovider/agent_e2e_test.go
```diff
@@ -0,0 +1,386 @@
...
+)
+
+func TestE2E_RunTextReturnsResponse(t *testing.T) {
+	client := newE2EClient(t)
+	canaryAgent := copilotprovider.NewAgent(client, copilotprovider.AgentConfig{
+		SessionConfig: restrictedSessionConfig(nil),
+	})
+	session := newE2ESession(t, canaryAgent, client)
+
+	ctx, cancel := context.WithTimeout(t.Context(), 2*time.Minute)
```
qmuntal: Don't hardcode timeouts. The testing framework already cancels the `t.Context()` when the test should time out. Adding another source of cancellation only adds noise. Same for subsequent WithTimeout Calls.

.github/workflows/test.yml
```diff
@@ -40,8 +40,41 @@ jobs:
...
+          cache-dependency-path: go.sum
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+
+      - name: Install Copilot CLI
+        shell: bash
+        run: npm install --global @github/copilot@1.0.75
```
qmuntal: Can we make this version handled by dependabot?

.github/workflows/live-examples.yml
```diff
@@ -0,0 +1,85 @@
...
+          cache-dependency-path: go.sum
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+
+      - name: Install Copilot CLI
+        shell: bash
+        run: npm install --global @github/copilot@1.0.75
```
qmuntal: Can we make this version handled by dependabot?

qmuntal (changes requested):

cmd/verifyexamples/examples.go
michelle-clayton-work: These examples have fixed inputs and stable output markers, so string checks are sufficient. Keeping ExpectedOutputDescription would require a Foundry verifier, adding an unrelated FOUNDRY_PROJECT_ENDPOINT dependency.

qmuntal (approved):
Thanks!

PR #1007, opened 2026-09-06
Detect changes in hosted content types in compaction contentEqual

PratikDhanave:
## Problem

`contentEqual` (`agent/compaction/equality.go`) is the change-detection predicate that `MessageIndex.Update` uses (via `contentsEqual`/`messageContentEqual`) to locate the already-processed prefix, and that `containsMessageByContent` uses for source attribution. Its `switch` handles text/reasoning/data/uri/error/function-call/function-result/hosted-file and returns `true` (**equal**) for everything else.

So two messages that differ only in a registered hosted content type — `MCPServerToolCallContent`, `MCPServerToolResultContent`, `CodeInterpreterToolCallContent`, `CodeInterpreterToolResultContent`, or `UsageContent` — are reported **equal**. `MessageIndex.Update` then treats the changed message as unchanged and does **not** rebuild, preserving stale group/exclusion state and mis-locating the prefix boundary. The existing `TestMessageIndexUpdate_MatchesKnownContentTypes` encodes the intent that any differing value must trigger a rebuild; the `default` violated it for these types.

## Fix

Add explicit equality cases for the hosted content types the framework actually produces, comparing their identity fields (recursing into `Outputs`/`Inputs` via the existing `contentsEqual`, and `reflect.DeepEqual` for `UsageDetails`). The `default` intentionally **stays** `return true` so a genuinely new, forward-compatible content type does not force an index rebuild — only the known hosted types get precise comparison.

## Test

Extends `TestMessageIndexUpdate_MatchesKnownContentTypes` with rows for each hosted type: identical values still preserve the index; differing `Arguments`/`Outputs`/`Inputs`/usage now correctly rebuild. The differing rows fail before the fix (treated as equal) and pass after.

### DIFF
--- a/agent/compaction/equality.go
+++ b/agent/compaction/equality.go
@@ -70,7 +70,29 @@ func contentEqual(left, right message.Content) bool {
 	case *message.HostedFileContent:
 		rightContent := right.(*message.HostedFileContent)
 		return leftContent.FileID == rightContent.FileID && leftContent.MediaType == rightContent.MediaType && leftContent.Name == rightContent.Name
+	case *message.MCPServerToolCallContent:
+		rightContent := right.(*message.MCPServerToolCallContent)
+		return leftContent.CallID == rightContent.CallID && leftContent.Name == rightContent.Name &&
+			leftContent.ServerName == rightContent.ServerName && leftContent.Arguments == rightContent.Arguments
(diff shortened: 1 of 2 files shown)


qmuntal:
Superseded by #1038, which includes this compaction equality coverage as part of the broader stable-content alignment.

PR #1008, opened 2026-09-06
Byte-account hosted tool-call/result content in compaction sizing

PratikDhanave:
## Problem

`computeContentByteCount` (`agent/compaction/index.go`) computes each group's `ByteCount` and, via `computeTokenCount`'s default branch (`… / 4`), the token estimate for non-text content. It accounts `FunctionResultContent` (`fmt.Sprint(Result)`) and `DataContent` (`len(Data)`), but `MCPServerToolResultContent` / `CodeInterpreterToolResultContent` (and the call variants) hit `default: return 0`.

So a `MCPServerToolResultContent`/`CodeInterpreterToolResultContent` carrying 10,000 bytes of `Outputs` contributes only ~2 bytes (just the CallID), whereas the equivalent `FunctionResultContent` contributes 10,004. Byte/token-based triggers (`TokensExceed`, and therefore `ContextWindowStrategy`) **under-count** history dominated by hosted code-interpreter / MCP outputs — precisely the large content that most needs compaction — so compaction can fail to fire when it should.

## Fix

Account the hosted tool-call/result types by recursing into their `Outputs`/`Inputs` (summing `computeContentByteCount`) and counting their identity strings, mirroring the function-call/result cases. `default` stays `0` for genuinely unknown types.

## Test

`TestMessageIndex_CountsHostedToolResultBytes` builds an MCP result and a code-interpreter result each carrying a 10k-byte `Outputs` text and asserts their `TotalByteCount` is at least that of the equivalent `FunctionResultContent`. Fails before the fix (2 bytes vs 10004), passes after.

### DIFF
--- a/agent/compaction/index.go
+++ b/agent/compaction/index.go
@@ -399,6 +399,26 @@ func computeContentByteCount(content message.Content) int {
 		return stringByteCount(typed.Message) + stringByteCount(typed.ErrorCode) + stringByteCount(typed.Details)
 	case *message.HostedFileContent:
 		return stringByteCount(typed.FileID) + stringByteCount(typed.MediaType) + stringByteCount(typed.Name)
+	case *message.MCPServerToolCallContent:
+		return stringByteCount(typed.CallID) + stringByteCount(typed.Name) + stringByteCount(typed.ServerName) + stringByteCount(typed.Arguments)
+	case *message.MCPServerToolResultContent:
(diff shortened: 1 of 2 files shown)


qmuntal:
@PratikDhanave please fix the Copilot review findings. Also, don't wait for me to say it, always do it on new PRs too,

qmuntal:
Superseded by #1038, which includes this recursive hosted-content byte accounting as part of the broader stable-content alignment.

PR #1058, opened 2026-09-13
Accept interface send type with a concrete implementing target in type validation

PratikDhanave:
## Problem

`sendTypeCompatibleWithInput` (`workflow/builder.go`, used by `validateTypeCompatibility` in `Build()`) checked:

```go
return outType == inType || outType.AssignableTo(inType) ||
    (inType.Kind() == reflect.Interface && outType.Implements(inType))
```

but **not** the symmetric `inType.AssignableTo(outType)`. The validator's contract is to reject an edge only when the source's send type-set and the target's accept type-set can never share a concrete value — an overlap relation that is **symmetric**. When the source declares an **interface** send type (e.g. `Shape`) and the target accepts a **concrete** type that implements it (e.g. `*Circle`), none of the clauses match, so `Build()` fails with:

```
type incompatibility between executors "source" -> "target": source sends [Shape] but target accepts [*Circle]
```

But the edge is runtime-valid: the source may emit a concrete `*Circle` (a legal member of its declared `Shape` set), and the router resolves it to the `*Circle` handler. The mirror direction (concrete send → interface target) already builds. This blocks the common `message.Content`-interface source → `*message.TextContent` concrete target pattern.

## Fix

Add the symmetric `inType.AssignableTo(outType)` clause. It does not over-accept — two unrelated concrete types are still non-assignable in both directions (verified: the existing `RejectsIncompatibleDeclaredSendTypes` test still passes). The removed third clause was already redundant with `outType.AssignableTo(inType)`.

## Test

`TestBuilder_Validation_TypeCompatibility_InterfaceSendConcreteTarget` builds a graph where the source declares an interface send type and the target accepts a concrete implementer, asserting `Build()` succeeds. Fails before the fix (rejected), passes after; existing compatibility tests remain green.


qmuntal (changes requested):
@PratikDhanave fix copilot findings. Also, do this without me having to say this on every existing and future PR please 😸 .

PratikDhanave:
Closing this — on reflection Copilot's finding is correct and my change was based on a faulty premise. `Context.SendMessage` resolves the source's declared send type via `DeclaredSendType`/`declaredSendTypeAssignableTo`, which deliberately excludes interface candidates ("interface send declarations match only that exact interface type"). So a source declaring `Sends(Shape)` that emits a concrete `*Circle` is rejected at runtime with `cannot send messages of type *Circle`. The build-time validation I relaxed was correctly mirroring that intentional runtime restriction, so accepting the edge at build time only defers the failure to runtime.

Making it genuinely work would mean also relaxing the runtime routing to route a concrete value under its declared interface type — a deliberate change to the documented routing model that's better decided as a design question than slipped in here. Withdrawing this rather than force an inconsistent half-fix. Sorry for the churn, and noted on getting ahead of the Copilot findings myself going forward.

PR #1062, opened 2026-09-13
Fall back to default output when WithOutputFrom is called with no agents

PratikDhanave:
## Problem

`outputDesignations.explicit()` returned `d != nil`, but `withOutputFrom` allocates a non-nil map even for **zero** agents:

```go
func (d outputDesignations) explicit() bool { return d != nil }

func (d outputDesignations) withOutputFrom(agents ...*agent.Agent) (outputDesignations, error) {
    if d == nil { d = make(outputDesignations) } // allocates even when len(agents)==0
    ...
}
```

So `WithOutputFrom()` / `WithIntermediateOutputFrom()` called with **no agents** — e.g. spreading a computed slice `builder.WithOutputFrom(outs...)` that happens to be empty — makes the designation set non-nil-but-empty. `explicit()` then reports `true`, so `applyDefaults()` (which designates the concurrent end / sequential output / group-chat host as the terminal output) is **skipped**, and the designation loop registers nothing. The built workflow has **zero output executors**: it runs but yields no result (only a `dead-end executors detected` WARN, no error). This affects the concurrent, sequential, and group-chat builders via the shared path.

## Fix

Treat an empty designation set as **non-explicit** (`len(d) != 0`), so an empty `WithOutputFrom()` falls back to the default terminal output instead of silently disabling all output. A designation with ≥1 agent is unaffected.

## Test

`TestConcurrentWorkflowBuilder_NoArgWithOutputFromFallsBackToDefault` builds a concurrent workflow with `.WithOutputFrom()` (no agents) and asserts `OutputExecutorIDs()` is non-empty. Fails before the fix (zero outputs), passes after.


qmuntal (changes requested):
Resolve consistency review findings.

PR #1063, opened 2026-09-13
Keep the original input under FreshContextPerIteration with ContinueWithMessages

PratikDhanave:
## Problem

`nextMessages` (`agent/harness/loop/loop.go`) returned the evaluator's explicit `ContinueWithMessages` messages **before** the `FreshContextPerIteration` branch:

```go
if len(evaluation.Messages) > 0 {
    cloned := cloneMessages(evaluation.Messages)
    return cloned, cloned   // returns explicit-only, bypassing fresh re-seed
}
if cfg.FreshContextPerIteration { ... rebuild from InitialMessages ... }
```

So when an evaluator returns `ContinueWithMessages(...)` while `FreshContextPerIteration` is enabled, the reinvocation gets **only** the explicit messages. Fresh mode also resets the session to a pristine snapshot taken before iteration 1 (which does not yet contain the original request), so the **original user input is lost from both the message list and the session**.

This contradicts the `FreshContextPerIteration` doc: *"restarts each reinvocation from the original input messages plus an aggregated feedback log …"*.

## Fix

Check `FreshContextPerIteration` first: always re-seed `InitialMessages`, and when the evaluator also supplied explicit messages, **compose** them on top of the fresh initial context (surfacing the explicit messages). Non-fresh mode is unchanged — explicit `ContinueWithMessages` is still sent verbatim (locked by `TestLoop_ContinueWithMessagesSendsMessagesVerbatim`).

## Test

`TestLoop_FreshContextPerIteration_ContinueWithMessagesKeepsInitial` runs a fresh-context loop whose evaluator returns `ContinueWithMessages(["explicit"])` and asserts the reinvocation still contains the original `"original"` input plus the explicit message. Fails before the fix (`[explicit]` only), passes after; the existing fresh-context + feedback test remains green.


qmuntal (changes requested):
Resolve consistency review findings.

PR #1068, opened 2026-09-14
Fix MCP tool discovery pagination

### LINKED ISSUE #1067: MCP tool discovery returns only first page of tools
Hi, I was testing an agent with MCP tools and noticed that `mcptool.ListTools` only returns tools from the first page.

My test server exposes 3 tools with one tool per page. The framework discovers only the first tool, even though the server return `NextCursor`. Because of this, tools from the next pages is not available to the agent.

### Expected behavior

`mcptool.ListTools` should read all pages and return all available tools

### What I checked

  - `session.ListTools(ctx, nil)` returns one page with a continuation cursor.
  - Using `session.Tools(ctx, nil)` discovers all 3 tools and allows calling the tool from the last page.
  - I also checked the .NET MCP SDK: its high-level `ListToolsAsync()` handles pagination automatically.

  ### Proposed fix

 Use `session.Tools(ctx, nil)` inside `mcptool.ListTools`, keeping the existing tool wrapping and normalized name collision checks.

I have a small fix prepared using `session.Tools(ctx, nil)`, with regression tests. Happy to open a PR if  this approach looks good.

gabisonia:
MCP servers can return tools across multiple pages, but `mcptool.ListTools` reads only the first page. This leaves tools from later pages unavailable to agents.

Use the SDK's `session.Tools(ctx, nil)` iterator to read all pages while preserving the existing tool wrappers and normalized name collision checks. If a later page fails, return the wrapped error without partial results.

Regression tests cover empty and single-page servers, discovery and invocation across multiple pages, normalized name collisions across pages, and later-page errors. The three pagination failure cases failed before the fix and pass with it.

Validation:

- `golangci-lint` v2.13.2 reported 0 issues.
- Changed Go files formatted with `gofmt`; `git diff --check` passed.

Fixes #1067


qmuntal (approved):
Thanks!

PR #1077, opened 2026-09-15
Stop tool invocation on request cancellation

### LINKED ISSUE #1076: Next tool still runs after request cancellation
I noticed that Go still calls the next tool after the request is cancelled.

The model requests two tools with sequential execution. Inside the first tool, I call `cancel()` and return `ctx.Err()` to simulate user pressing Stop. The second tool still starts with an already cancelled context.

I tested the same with .NET `FunctionInvokingChatClient`, and it stops before the second tool.

gabisonia:
Fixes #1076

When a request is cancelled during a tool call, the invocation loop can treat cancellation as a recoverable tool error and start the next tool. This change checks the request context before provider/tool calls and after tool execution, so cancellation stops further work and is returned to the caller.

Ordinary tool errors and cancellation of a tool's separate context keep their existing handling. Already-running concurrent tools still need to observe their context; the loop waits for them before returning cancellation.

Added regression tests for cancellation before and during tool calls, deadlines, concurrent calls, cancellation after tool results, and independent tool errors.

Validation:
- New cancellation tests fail on the original code and pass with the fix.
- golangci-lint v2.13.2 reported no issues.
- A live OpenAI Responses test with `gpt-4.1-mini` confirmed that Go stops before the second tool and makes no follow-up HTTP transport attempt. The equivalent .NET `FunctionInvokingChatClient` test also stops before the second tool.


agent/harness/toolautocall/autocall_cancellation_test.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Microsoft. All rights reserved.
```
qmuntal: Don't create a new file for these tests.

agent/harness/toolautocall/autocall.go
```diff
@@ -1056,11 +1064,18 @@ func (f *autocall) processFunctionCalls(ctx context.Context, tools map[string]to
...
+			return nil, errCount, err
+		}
 		results = parallelResults
 	} else {
 		// Invoke each function serially.
 		for _, fc := range funcCalls {
 			result := f.processFunctionCall(ctx, tools, fc)
+			// Request cancellation must bypass the recoverable tool-error path.
+			if err := ctx.Err(); err != nil {
+				return nil, errCount, err
```
qmuntal, line 1092: Could we preserve the original tool error when cancellation is detected, both here and in the concurrent branch?

If a tool cancels the request and returns an ordinary error, this returns `context.Canceled` instead and discards the actual failure. In Microsoft.Extensions.AI 10.10.0, the [.NET exception filter](https://github.com/dotnet/extensions/blob/294665a5ec4eceaa245388cfb70883ea01881081/src/Libraries/Microsoft.Extensions.AI/Common/FunctionInvocationProcessor.cs#L131-L149) only captures exceptions while the request token is not canceled. Once cancellation is requested, the original exception propagates unchanged. This difference was reproduced with a fake provider and tool in both streaming and non-streaming calls.

Please preserve the original tool failure when present, fall back to `ctx.Err()` when there is no tool error.

qmuntal (changes requested):

agent/harness/toolautocall/autocall_cancellation_test.go
gabisonia: Sure, I’ll move these tests into autocall_test.go

agent/harness/toolautocall/autocall.go
gabisonia: Thanks, I’ll keep the original error and fall back to ctx.Err() if there isn’t one.

qmuntal (approved):

PR #1088, opened 2026-09-16
Preserve raw JSON in tool result compaction

### LINKED ISSUE #1087: Tool result compaction converts raw JSON into byte numbers
`ToolResultStrategy` turns `json.RawMessage` results into byte numbers in the summary.

For example, ``json.RawMessage(`{"status":"shipped"}`)`` becomes `[123 34 115 ...]` instead of readable JSON.

`DefaultToolCallFormatter` uses `fmt.Sprint` on the result. Returning a plain string works fine.

Also tested .NET with `JsonElement` and it keeps the JSON text.

gabisonia:
Fixes #1087

`ToolResultStrategy` currently formats `json.RawMessage` results as byte numbers. Preserve the JSON text in the summary by converting raw messages directly to strings, while keeping the existing formatting for other result types.

Added regression coverage in the existing compaction test file for raw JSON objects, arrays, null, empty/nil values, and ordinary string and number results. The raw JSON cases fail before the fix and pass afterward.

Validation: `go test -race -shuffle=on ./...`, golangci-lint (zero issues), and the local reproduction sample against the fix.


qmuntal (approved):
Thanks!

Issue #949, opened 2026-08-31
Expose tool invocation identity through delegating-agent middleware


hlgone:
## Summary

Expose the current tool invocation identity, including the function call ID, through the delegating-agent middleware design proposed by the maintainers in #638.

This should preserve the idiomatic `FuncTool.Call(context.Context, args)` API introduced by #205. The goal is not to restore the removed custom `tool.Context`, but to provide a supported per-invocation interception contract for policy, approval, audit, and gateway integrations.

## Motivation

The automatic tool-call loop already has the originating `message.FunctionCallContent` and uses its `CallID` to correlate approval requests, function results, and OpenTelemetry spans. However, application middleware and tool wrappers do not currently have a supported public contract for that identity.

Enterprise tool gateways commonly need to bind:

- a policy decision to the exact logical tool call;
- an approval request and its resumed execution;
- normalized argument digests and consume-once records;
- audit events and function results;
- parallel calls to separate correlation records.

Passing this identity as a model-generated tool argument is not appropriate because it mixes framework metadata with model-controlled business arguments.

## Existing direction

PR #638 demonstrated the need for per-function interception. It was closed with maintainer guidance that the capability should be implemented as an `agent.Middleware` that replaces real tools with lifecycle-aware wrappers, similar to the .NET delegating-agent design.

This proposal follows that direction.

## Proposed contract

A delegating-agent middleware should be able to observe a per-invocation context containing at least:

- the invoked tool;
- the original function-call content or an explicit `CallID`;
- the normalized invocation arguments;
- the invocation result or error;
- the current `context.Context`.

The exact API shape can follow the wrapper approach preferred in #638. Two possible non-exclusive surfaces are:

1. a public function-invocation context passed to the wrapper callback;
2. a typed accessor such as `tool.InvocationFromContext(ctx)` when the wrapped `FuncTool` itself needs the metadata.

The second option should keep using standard `context.Context`; it should not reintroduce a custom context parameter.

## Semantics to document

- The call ID identifies one logical tool invocation.
- The same approved invocation should expose the same call ID when execution resumes.
- Parallel invocations must expose distinct IDs.
- Provider behavior when no call ID is supplied should be documented, including whether the framework creates a synthetic ID.
- The call ID is a correlation identifier, not a cross-run idempotency key.
- Authorization must not rely on the call ID alone.

## Suggested acceptance tests

1. A delegating middleware observes the call ID before invoking a normal tool.
[...]

PR #1093, opened 2026-09-16
Add function invocation middleware with call identity

gabisonia:
Related to #949

Adds function-invocation middleware so applications can inspect the tool, call ID, arguments, result, and error for each call. It follows the approach suggested in #638: `agent.FunctionInvocationMiddleware` implements `agent.Middleware` and wraps function tools. Automatic tool execution supplies the call ID through `context.Context`; `FuncTool.Call(ctx, args)` stays unchanged.

```go
audit := agent.FunctionInvocationMiddleware(func(
    ctx context.Context,
    invocation *agent.FunctionInvocationContext,
    next agent.FunctionInvocationFunc,
) (any, error) {
    log.Printf("Starting %s: %s", invocation.Function.Name(), invocation.CallID)
    result, err := next(ctx, invocation)
    log.Printf("Finished %s: %v", invocation.CallID, err)
    return result, err
})
```

Register the callback in `ProviderConfig.Middlewares` before `toolautocall`. Callbacks can replace arguments, inspect or replace returned results/errors, or skip the underlying tool by not calling `next`. Skipping one invocation does not stop the whole agent loop. Tool schemas and approval requirements are preserved, and callbacks run after approval. `tool.InvocationFromContext` also exposes the ID inside tool handlers.

Opening as a draft to discuss the API shape and argument contract. #949 mentions normalized arguments; this implementation exposes raw JSON to match Go's existing tool API. Should normalization belong here or remain in the tool/application? The full issue contract still needs agreement.

The middleware wraps tools present in agent options at its position in the pipeline. An internal run option applies the same wrappers to separately configured `AdditionalTools`, without adding them to provider requests. Request tools retain precedence. Missing call IDs remain empty, and IDs are for correlation, not authorization or cross-run idempotency.

Validation: full race-enabled suite (`go test -race -shuffle=on ./...`), targeted callback/identity/approval/cancellation race tests, and golangci-lint (zero issues). Tests are in existing files and cover concurrent calls, approval/session restoration, context propagation, callback ordering, argument/result changes, errors, and repeated runs. An offline sample was also run locally.


gabisonia:
@qmuntal could you take a look at the API when you have time? I kept arguments as raw JSON to match `FuncTool.Call`. Would you prefer normalized arguments here?

agent/middleware.go
```diff
@@ -38,6 +40,98 @@ func (mf MiddlewareFunc) Run(next RunFunc, ctx context.Context, messages []*mess
...
+	Function tool.FuncTool
+
+	// CallID identifies the originating function call. Treat this field as read-only;
+	// changing it does not change the ID used by the framework for its result.
+	// It is empty when the provider supplied no ID or there is no invocation context.
+	CallID string
+
+	// Arguments contains the raw JSON passed to the tool. Middleware may replace
+	// it before calling next. The tool retains responsibility for input validation.
+	Arguments string
```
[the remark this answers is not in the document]
gabisonia: Clarified in 92ed243: raw JSON matches FuncTool.Call, with validation handled by the tool.

agent/middleware.go
```diff
@@ -38,6 +40,102 @@ func (mf MiddlewareFunc) Run(next RunFunc, ctx context.Context, messages []*mess
...
+// after components that supply tools. For tools from context providers, use
+// [ProviderConfig.Middlewares]. Function tools configured separately on the
+// automatic tool-call middleware are also wrapped, without adding them to provider requests.
+// Approval requirements and tool schemas are preserved. Callbacks run only when
+// the tool is invoked, including after approval, not when approval is requested.
+//
+// Multiple callbacks execute in registration order, with the first outermost.
+// Each call gets its own FunctionInvocationContext; callbacks must synchronize
+// shared application state if tools can execute concurrently.
+type FunctionInvocationMiddleware func(ctx context.Context, invocation *FunctionInvocationContext, next FunctionInvocationFunc) (any, error)
```
qmuntal, line 82: `next` should be the first parameter, matching `Middleware` ordering.

agent/middleware.go
```diff
@@ -38,6 +40,102 @@ func (mf MiddlewareFunc) Run(next RunFunc, ctx context.Context, messages []*mess
...
+	// receives it before any validation or normalization performed by the
+	// underlying tool.
+	//
+	// Middleware may replace it before calling next. To inspect or change
+	// individual fields, decode the JSON and encode any changes back into Arguments.
+	Arguments string
+}
+
+// FunctionInvocationFunc invokes the next function middleware or underlying tool.
+type FunctionInvocationFunc func(context.Context, *FunctionInvocationContext) (any, error)
```
qmuntal, line 64: `FunctionInvocationFunc` is simple enough, and not important enough, to deserve a type for now.

agent/middleware.go
```diff
@@ -38,6 +40,102 @@ func (mf MiddlewareFunc) Run(next RunFunc, ctx context.Context, messages []*mess
...
+
+func (t *functionInvocationTool) ApprovalRequired() bool {
+	approval, ok := t.FuncTool.(tool.ApprovalRequiredTool)
+	return ok && approval.ApprovalRequired()
+}
+
+func (t *functionInvocationTool) Call(ctx context.Context, args string) (any, error) {
+	identity, _ := tool.InvocationFromContext(ctx)
+	invocation := &FunctionInvocationContext{Function: t.FuncTool, CallID: identity.CallID, Arguments: args}
+	var next FunctionInvocationFunc = func(ctx context.Context, invocation *FunctionInvocationContext) (any, error) {
```
qmuntal, line 127:
```suggestion
	next := func(ctx context.Context, invocation *FunctionInvocationContext) (any, error) {
```

tool/tool.go
```diff
@@ -83,9 +85,33 @@ type FuncTool interface {
...
+//
+// The identity is scoped to the invocation context, including contexts derived by
+// wrappers, and is independent for parallel calls. A direct Call with an unrelated
+// context has no invocation identity. Direct calls made with an existing invocation
+// context inherit that identity; they do not create a new logical function call.
+func InvocationFromContext(ctx context.Context) (Invocation, bool) {
+	callID, ok := toolcontext.CallIDFromContext(ctx)
+	return Invocation{CallID: callID}, ok
+}
+
```
qmuntal, line 114: Can we keep these two symbols unexported in an internal package? AFAIU the intended user is `FunctionInvocationMiddleware `, no?

internal/toolcontext/context.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// Package toolcontext carries tool invocation metadata across framework packages.
+package toolcontext
```
qmuntal, line 4: Don't like having two internal packages for tool helpers. consolidate into one (I like the `toolmiddleware` name a bit more)

qmuntal (changes requested):
Like the overall approach 😄

gabisonia:
Thanks for the review! Updated the API and moved the internal helpers into one package. Also merged latest main.

agent/agent_test.go
```diff
@@ -21,6 +22,106 @@ type stubTool struct {
...
+					if result != want {
+						t.Errorf("result = %v, want %q", result, want)
+					}
+					yield(&agent.ResponseUpdate{Role: message.RoleAssistant, Contents: message.Contents{&message.TextContent{Text: "done"}}}, nil)
+				}
+			}
+			var nilMiddleware agent.FunctionInvocationMiddleware
+			a := agent.New(agent.ProviderConfig{Run: run, Middlewares: []agent.Middleware{first, nilMiddleware, second}}, agent.Config{
+				ContextProviders: []agent.ContextProvider{provider},
+			})
```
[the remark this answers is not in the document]
gabisonia: This already uses ProviderConfig.Middlewares. The test also passes with -race.

qmuntal (changes requested):
1. **Context-provider tools bypass middleware with the standard constructors.**  
   `Config.Middlewares` runs `agent.go:94-95`, but the `chat.go:80-98` doesn’t expose `ProviderConfig.Middlewares`, where the PR recommends registering it.

   Reproduced with `NewChatCompletionsAgent`: middleware blocked a configured tool, but the same tool supplied by a context provider executed with **zero middleware callbacks**. There needs to be an accessible registration point after context collection, or deferred wrapping of newly added tools.

2. **Copilot tool calls lose their call identity.**  
   The `copilot.go:493-502` calls `FuncTool.Call` directly, without attaching the SDK’s `ToolCallID`. The middleware executes, but receives an empty `CallID` despite the SDK providing one. This path should attach the same internal invocation metadata as `toolautocall`.

qmuntal (approved):
I’d avoid adding `ProviderMiddlewares`. Could we use `Config.FunctionMiddlewares` and apply it around each tool call instead? That would cover context-provider tools and `AdditionalTools` without requiring users to get the middleware ordering right.

.NET and Python also separate function middleware registration from execution.

gabisonia:
Makes sense, I’ll switch to Config.FunctionMiddlewares so users don’t need to worry about ordering.

qmuntal (approved):
Thanks!

PR #1113, opened 2026-09-20
Preserve MCP array output schemas and structured results

gabisonia:
`mcptool.AddTool` currently drops the output schema for array results and returns them only as JSON text. This adds an output schema and structured content for arrays and nullable arrays.

For example, a tool returning `["ORD-1042","ORD-1043"]` now also provides `structuredContent: {"result":["ORD-1042","ORD-1043"]}`. The output schema describes that wrapper, while the existing text response stays unchanged, including formatting in `json.RawMessage` values.

The object wrapper supports older MCP clients and is used consistently across protocol versions. This matches .NET's legacy structured-output shape; modern .NET clients receive a natural array instead. Keeping the original text here avoids breaking clients that already parse it as an array.

Local schema references are relocated without modifying the original schema or literal default/example data. Serialization failures return tool errors. Explicit MCP/framework content results retain their existing behavior.

Validation:
- Regression tests in the existing `mcp_test.go` cover three protocol versions, empty/nil arrays, schema references and IDs, schema immutability, raw JSON text, and serialization errors.
- `golangci-lint` v2.13.2: 0 issues.
- Formatting and `git diff --check` passed.


PratikDhanave:
Went through this carefully (I was recently in `mcptool/mcp.go` for #1135, so it was fresh). The wrap-array-output-in-`{ "result": ... }` approach is the right workaround for MCPs object-only `outputSchema` requirement, and the `$ref` relocation is more rigorous than I expected — nicely done. Some specific things I checked and think are correct:

- **Local `$ref` rewrite** (`"#/properties/result" + ref[1:]`) correctly handles both the root self-ref (`#` → `#/properties/result`) and `#/$defs/...` (→ `#/properties/result/$defs/...`), and since the whole array schema (including its `$defs`) moves under `properties.result`, the targets travel with it. Good that you rewrite from the raw (still-escaped) `ref[1:]` and only unescape for the prefix test.
- **`$id` short-circuit**: returning early when a subschema declares `$id` is the correct call — that subschema establishes its own base URI, so its internal refs must *not* be rebased to the document root. 👍
- **`UseNumber()`** to avoid float-clobbering large integers in the copied schema is a nice touch (and I see the large-number test covering it).

Two questions / possible edge cases:

1. **Array output schema + explicit content result.** When `wrapOutput` is true but the tool returns `mcp.Content` / `message.Content` (the `default`-excluded cases), the result keeps its existing conversion but the advertised `OutputSchema` is the wrapped `{result: array}`. Thats a schema/result shape mismatch for that (admittedly unusual) combination. Is that intended to just fall back silently, or worth a doc note that an array return schema assumes a structured (non-content) result?

2. **`$recursiveRef`.** You handle `$ref` and `$dynamicRef`; older drafts use `$recursiveRef`. Probably out of scope if `jsonschema-go` only ever emits `$dynamicRef`, but worth confirming thats the only dynamic form the generator produces.

Neither is blocking — this looks solid and well-tested. Nice work.

gabisonia:
Thanks for checking! Added the doc note. The jsonschema-go version we use doesn’t generate `$recursiveRef`, so I’m leaving that outside this fix.

qmuntal (approved):
Thanks.

PR #1127, opened 2026-09-20
Avoid messages for response-only updates

llirik0:
## Summary

Prevent response-level metadata updates from creating an empty assistant message when a response is converted with `ToUpdates` and collected again.

`Response.Update` now creates or mutates message state only when an update carries message data. Continuation tokens and response-level additional properties still round-trip normally.

## Testing

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run`


llirik0:
> Kirill (@llirik0) please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

PratikDhanave:
Reviewed this closely since it touches the same `Response.Update`/`ToUpdates` seam I was in for #1133. The core change is right and matches M.E.AI `ChatResponseExtensions.ProcessUpdate`: a response-metadata-only update should not materialize an empty message. Tests are a good round-trip shape. A few notes:

**Composition with #1133 (worth the maintainers knowing before merge order is decided).** #1133 scopes `AdditionalProperties` in `Response.Update` to the message when the update carries a `MessageID` and to the response otherwise. These two PRs are complementary and both needed:
- *Metadata-only update (no MessageID)*: this PR already skips message creation, so `AdditionalProperties` only reaches the response — which happens to give the same result #1133 does for that case.
- *Update with a MessageID + AdditionalProperties*: this PR still enters the message block, where today the map is copied to **both** the message and the response (the double-copy #1133 fixes). This PR does not address that case.

So they dont conflict, but whichever merges second will want a trivial rebase in the same block. No action needed here beyond awareness.

**Question — `RawRepresentation != nil` in the guard.** Including it means a provider update that carries only a raw representation (no id/role/contents) will still create an empty-content message. Within `ToUpdates` this cant happen (the extra response-level update has `RawRepresentation == nil`), so the round-trip tests pass — but for a live streaming provider that emits a final raw-only event, is materializing a message the intended behavior, or should raw-only also be treated as response-level? Might be worth a comment on the guard either way.

**Question — the `ToUpdates` role default (`msg.Role == "" -> RoleAssistant`).** This is a separate behavior tweak bundled into the same PR. It looks reasonable, but does it mirror what .NET does for an unroled message on `ToChatResponseUpdates`, and is it needed for the metadata-message case specifically? A one-line rationale in the code or PR body would help future readers see why it rides along.

Nothing blocking from my perspective — the direction is correct and well-tested. 👍

